Issue #96, opened 2021-11-29
feat: update play/pause/stop functionality per new design


justinkambic:
Related to https://github.com/elastic/synthetics-recorder/issues/70.

Per designs in the linked meta issue, we should implement a new flow for play/pause/stop in the recorder's flow.

- [x] Start button will now become Pause during a run
- [x] No Stop button, stop button is now "start over", which will re-launch a new session.
  - for now, this will kill the existing chromium process, and focus on the URL input. Eventually we will have a "start" page that consists of just the URL input and a start button.

More info/design on the parent issue.

justinkambic:
I think it'll be better to update this functionality first and then treat the addition of a startup screen as an isolated change.

[I also think](https://github.com/elastic/synthetics-recorder/issues/70#issuecomment-983758099) it might be better if, once they're past the start screen, we don't return them there when they start over. That's just my opinion of course.

awahab07:
**Post FF Testing**

LGTM
"Start recording" button becomes "Pause" during a recording session. 
There's no "Stop" button, rather during a recording session, there's a "Start over" button.
Before starting over, "Start over" button confirms from the user to delete all recorded steps and start over.

https://user-images.githubusercontent.com/2748376/148784815-806dc195-9ad2-464d-926a-8ff0bf4de1fc.mov

PR #97, opened 2021-11-29
Update play/pause functionality to match designs

justinkambic:
Resolves https://github.com/elastic/synthetics-recorder/issues/96.

Modifies the Play/Pause functionality to share the same button. Updates stop functionality to depend on chromium process ending. Adds new "start over" abort functionality.

As pictured below, the user can now "start over", which will cause the main process to kill the recording session and emit an empty set of steps for the UI to display. It also causes the front-end to re-focus the URL input. They can then modify the URL if they so choose and start a new recording session. There's a modal that requires confirmation before taking the destructive action of deleting all of the steps they've previously recorded.

### DIFF
--- a/electron/execution.js
+++ b/electron/execution.js
@@ -111,17 +111,7 @@ async function recordJourneys(data, browserWindow) {
       await browser.close().catch({});
     };
 
-    // indicates the client wants to quit the session and keep
-    // the current state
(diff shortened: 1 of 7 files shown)


justinkambic:
Closing this in favor of https://github.com/elastic/synthetics-recorder/pull/105 for the reasons explained there.

PR #107, opened 2021-12-03
feat: Implement updated header UI

### LINKED ISSUE #106: feat: Update header to match new design
We have a new design for the recorder's header:

image

Update the UI to reflect this. It will ensure we don't have any buttons hidden at the bottom of the UI, and make the body of the recorder less cluttered in general.

justinkambic:
Resolves #106.

First step at implementing the latest recorder designs. This updates the header to match the styling, and migrates the buttons and functionality from elsewhere in the app to be in the header.

We can hold off on merging this until the follow-up PR adding the updated body is ready. My goal with this dedicated PR is to limit the scope and review burden.

### DIFF
--- a/electron/execution.js
+++ b/electron/execution.js
@@ -126,7 +126,7 @@ async function onTest(data) {
     succeeded: 0,
     failed: 0,
     skipped: 0,
-    journeys: {},
+    journey: {},
   };
 
   const parseOrSkip = chunk => {
@@ -141,13 +141,17 @@ async function onTest(data) {
     switch (parsed.type) {
       case "journey/start": {
         const { journey } = parsed;
-        result.journeys[journey.name] = { status: "succeeded", steps: [] };
+        result.journey = {
+          status: "succeeded",
+          steps: [],
+          type: journey.name,
+        };
         break;
       }
       case "step/end": {
-        const { journey, step, error } = parsed;
+        const { step, error } = parsed;
         result[step.status]++;
-        result.journeys[journey.name].steps.push({
+        result.journey.steps.push({
           name: step.name,
           status: step.status,
           error,
@@ -157,7 +161,7 @@ async function onTest(data) {
       }
       case "journey/end": {
         const { journey } = parsed;
-        result.journeys[journey.name].status = journey.status;
+        result.journey.status = journey.status;
         break;
       }
     }
--- a/src/common/shared.ts
+++ b/src/common/shared.ts
@@ -62,6 +62,11 @@ export const COMMAND_SELECTOR_OPTIONS = [
   },
 ];
 
+export const SYNTHETICS_DISCUSS_FORUM_URL =
+  "https://forms.gle/PzVtYoExfqQ9UMkY6";
+
+export const SMALL_SCREEN_BREAKPOINT = 850;
+
 export function performSelectorLookup(
   onSelectorChange: Setter<string | undefined>
 ) {
@@ -109,9 +114,6 @@ export function updateAction(
   });
 }
 
-export const SYNTHETICS_DISCUSS_FORUM_URL =
-  "https://forms.gle/PzVtYoExfqQ9UMkY6";
-
 export async function getCodeForResult(
   steps: ActionContext[][],
   journey: Journey | undefined
@@ -130,5 +132,3 @@ export async function getCodeForResult(
     journey.type
   );
 }
-
-export const SMALL_SCREEN_BREAKPOINT = 850;
--- a/src/common/shared.ts
+++ b/src/common/shared.ts
@@ -23,13 +23,7 @@ THE SOFTWARE.
 */
 
 import React from "react";
-import type {
-  ActionContext,
-  Journey,
-  JourneyStep,
-  JourneyType,
-  Setter,
-} from "./types";
+import type { ActionContext, Journey, JourneyType, Setter } from "./types";
 
 const { ipcRenderer: ipc } = window.require("electron-better-ipc");
 
@@ -118,45 +112,23 @@ export function updateAction(
 export const SYNTHETICS_DISCUSS_FORUM_URL =
   "https://forms.gle/PzVtYoExfqQ9UMkY6";
 
-export function combineResultJourneys(journey: Journey) {
-  const journeyArr = [];
-  if (journey.inline) {
-    journeyArr.push(journey.inline);
-  }
-  if (journey.suite) {
-    journeyArr.push(journey.suite);
-  }
-  return journeyArr;
-}
-
-function getStepActions(step: JourneyStep, currentActions: ActionContext[][]) {
-  if (!step.name) return;
-  for (let i = 0; i < currentActions.length; i++) {
-    if (
-      currentActions[i].length > 0 &&
-      currentActions[i][0].title === step.name
-    ) {
-      return currentActions[i];
-    }
-  }
-}
-
 export async function getCodeForResult(
-  actions: ActionContext[][],
-  journey: Journey | undefined,
-  type: string
-) {
+  steps: ActionContext[][],
+  journey: Journey | undefined
+): Promise<string> {
   if (!journey) return "";
-  const journeyArr = combineResultJourneys(journey);
-  const stepActions = journeyArr
-    .map(({ steps }) => {
-      for (const step of steps) {
-        return getStepActions(step, actions) ?? null;
-      }
-      return null;
-    })
-    .filter(f => f !== null);
+  const journeyStepNames = new Set(journey.steps.map(({ name }) => name));
 
-  // @ts-expect-error null elements are filtered out
-  return await getCodeFromActions(stepActions ?? [], type);
+  return await getCodeFromActions(
+    steps.filter(
+      step =>
+        step !== null &&
+        step.length > 0 &&
+        step[0].title &&
+        journeyStepNames.has(step[0].title)
+    ),
+    journey.type
+  );
 }
+
+export const SMALL_SCREEN_BREAKPOINT = 850;
--- a/src/common/shared.ts
+++ b/src/common/shared.ts
@@ -23,7 +23,13 @@ THE SOFTWARE.
 */
 
 import React from "react";
-import type { ActionContext, JourneyType, Setter } from "./types";
+import type {
+  ActionContext,
+  Journey,
+  JourneyStep,
+  JourneyType,
+  Setter,
+} from "./types";
 
 const { ipcRenderer: ipc } = window.require("electron-better-ipc");
 
@@ -108,3 +114,50 @@ export function updateAction(
     });
   });
 }
+
+export const SYNTHETICS_DISCUSS_FORUM_URL =
+  "https://forms.gle/PzVtYoExfqQ9UMkY6";
+
+export function combineResultJourneys(journey: Journey) {
+  const journeyArr = [];
+  if (journey.inline) {
+    journeyArr.push(journey.inline);
+  }
+  if (journey.suite) {
+    journeyArr.push(journey.suite);
+  }
+  return journeyArr;
+}
+
+function getStepActions(step: JourneyStep, currentActions: ActionContext[][]) {
+  if (!step.name) return;
+  for (let i = 0; i < currentActions.length; i++) {
+    if (
+      currentActions[i].length > 0 &&
+      currentActions[i][0].title === step.name
+    ) {
+      return currentActions[i];
+    }
+  }
+}
+
+export async function getCodeForResult(
+  actions: ActionContext[][],
+  journeys: Journey | undefined,
+  type: string
+) {
+  if (!journeys) return "";
+  const journeyArr = combineResultJourneys(journeys);
+  const stepActions = journeyArr
+    .map(({ steps }) => {
+      for (const step of steps) {
+        return getStepActions(step, actions) ?? null;
+      }
+      return null;
+    })
+    .filter(f => f !== null);
+
+  // @ts-expect-error null elements are filtered out
+  return await getCodeFromActions(stepActions ?? [], type);
+}
+
--- a/src/common/shared.ts
+++ b/src/common/shared.ts
@@ -143,11 +143,11 @@ function getStepActions(step: JourneyStep, currentActions: ActionContext[][]) {
 
 export async function getCodeForResult(
   actions: ActionContext[][],
-  journeys: Journey | undefined,
+  journey: Journey | undefined,
   type: string
 ) {
-  if (!journeys) return "";
-  const journeyArr = combineResultJourneys(journeys);
+  if (!journey) return "";
+  const journeyArr = combineResultJourneys(journey);
   const stepActions = journeyArr
     .map(({ steps }) => {
       for (const step of steps) {
--- a/src/common/shared.ts
+++ b/src/common/shared.ts
@@ -160,4 +160,3 @@ export async function getCodeForResult(
   // @ts-expect-error null elements are filtered out
   return await getCodeFromActions(stepActions ?? [], type);
 }
-
--- a/src/components/Header/HeaderControls.tsx
+++ b/src/components/Header/HeaderControls.tsx
@@ -35,7 +35,7 @@ import { TestContext } from "../../contexts/TestContext";
 import { ControlButton } from "../ControlButton";
 import { SaveCodeButton } from "../ExportScriptButton";
 import { TestButton } from "../TestButton";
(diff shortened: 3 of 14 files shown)


src/components/Header/HeaderControls.tsx
```diff
@@ -0,0 +1,118 @@
...
+  return (
+    <EuiFlexGroup
+      alignItems="center"
+      gutterSize="m"
+      style={{
+        backgroundColor: euiTheme.colors.lightestShade,
+        borderBottom: euiTheme.border.thin,
+        margin: "0px 0px 4px 0px",
+        padding: 8,
+      }}
```
lucasfcosta, line 75: Should we encapsulate this into a `styles` map and refer to it within the component instead of inlining? Nbd though, just asking what we usually do in these cases.

src/components/Header/HeaderControls.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const { onTest } = useContext(TestContext);
+  return (
+    <EuiFlexGroup
+      alignItems="center"
+      gutterSize="m"
+      style={{
+        backgroundColor: euiTheme.colors.lightestShade,
+        borderBottom: euiTheme.border.thin,
+        margin: "0px 0px 4px 0px",
+        padding: 8,
```
lucasfcosta, line 74: Don't we need a unit for padding here?

src/components/StepsMonitor.tsx
```diff
@@ -181,39 +125,38 @@ function CodeFlyout({
...
+      <EuiSpacer />
+
+      {isFlyoutVisible && (
+        <CodeFlyout
+          actions={actions}
+          code={code}
+          setCode={setCode}
+          setIsFlyoutVisible={setIsFlyoutVisible}
+          setType={setType}
+          type={type}
```
lucasfcosta, line 157: [Nit] Should we manage the type selection within the `CodeFlyout` itself given it's not used on the this component?

lucasfcosta (changes requested):
Excellent work (as usual), thank you @justinkambic!

Besides the code-comments, here is some general feedback as I was end-to-end testing this PR:

* It doesn't matter how much I expand the recorder window, it still shows this gap between the right side of the header and the window border.
    Screenshot 2021-12-13 at 15 59 36
021-12-13 at 15 28 16](https://user-images.githubusercontent.com/6868147/145845789-da9fe563-0c8c-4d2b-bb77-6b063a48b5b9.png)
* [Nit] The designs show the header as being fully white, while in the recorder it looks a bit grayish (I used the Mac's colour picker to confirm we're not using `#FFF` for the header's colour)
   Screenshot 2021-12-13 at 16 03 21
* When I "pause" the recording, the "view code" button looks disabled, but I can still click it and open the flyout (I don't know if that's the expected behaviour, but I've also added that as a comment on the designs so that @liciavale can review). I assume it could be okay.
* Currently, when I "pause" the recording, I can still click "replay steps", which is _not_ disabled, but that button doesn't do anything.
   Screenshot 2021-12-13 at 16 10 46
* The colour of the status dot for the "recording paused" status doesn't seem to match the designs. In the screenshot below, the big dot is the one from the designs, while the small one is the actual colour in the recorder app.
   Screenshot 2021-12-13 at 16 11 24
* [There's another code comment for this] Shouldn't we disable the recording button if the user hasn't entered a journey to record? I assume we don't want users to ever start from `about:blank`, is that correct? (CC @liciavale)

Please feel free to dismiss the comments you consider to be purely stylistic or just let me know if you think there could be alternatives that are similar to what I suggested. I'm not totally sure about how we usually like to style code in terms of imperative vs. functional paradigms, so I'm mostly going for what feels shorter/natural to me, which is somewhat personal, but I think may indeed improve readability objectively sometimes.

justinkambic:
@lucasfcosta thanks for such detailed feedback. I'm going to answer your top-level review mentions in this comment, and then I'll reply to/resolve the granular code comments inline.

- > It doesn't matter how much I expand the recorder window, it still shows this gap between the right side of the header and the window border.

Interesting.. this isn't happening for me locally:

image

- > [Nit] The designs show the header as being fully white, while in the recorder it looks a bit grayish (I used the Mac's colour picker to confirm we're not using #FFF for the header's colour)

  The background color is inherited from the EUI component's default, but I will update this to use `euiEmptyShade`, which I believe is #FFF for the light theme.

- > When I "pause" the recording, the "view code" button looks disabled, but I can still click it and open the flyout (I don't know if that's the expected behaviour, but I've also added that as a comment on the designs so that @liciavale can review). I assume it could be okay.

  This is a potential design issue. The button is simply styled with the `text` color, I'm assuming it's intended to make it understated. AFAIK this doesn't change in any case, so it's possible we will want to re-style it. Perhaps as an `empty` button with a `secondary` color, or some other variation, to not make it appear disabled.

- > The colour of the status dot for the "recording paused" status doesn't seem to match the designs. In the screenshot below, the big dot is the one from the designs, while the small one is the actual colour in the recorder app.

  The color isn't taken directly from the design - all of the colors in the recorder are using the EUI theme's color exports. It's an important best practice to avoid deviating from the EUI colors without good reason, because the EUI team maintains them in the larger context of the overall theme. When they need to change colors, all the downstream elements benefit without need for an implicit code change.

  It's also important to follow this from an accessibility standpoint where possible, because they sometimes need to tweak the palette when issues arise. @liciavale LMK if you want that specific color for the warning state.

- > Shouldn't we disable the recording button if the user hasn't entered a journey to record? I assume we don't want users to ever start from about:blank, is that correct? (CC @liciavale)

  I'd be fine with this constraint, but this is also the way the recorder already works today. Curious if @vigneshshanmugam has any concerns about them opening up an empty chromium.

src/common/shared.ts
```diff
@@ -108,3 +114,49 @@ export function updateAction(
...
+      currentActions[i][0].title === step.name
+    ) {
+      return currentActions[i];
+    }
+  }
+}
+
+export async function getCodeForResult(
+  actions: ActionContext[][],
+  journeys: Journey | undefined,
```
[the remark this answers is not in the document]
justinkambic: You'll find a bunch of code like this - part of the growing pains of the project. We went from pure JS without established naming conventions in the prototype phase and we're working toward a mature TS codebase, so there are some problematic naming choices at the moment. I'm working on fixing this a bit downstream from this branch, but yeah it's a single `journey`, so we'll rename it to that 😄

src/components/Header/HeaderControls.tsx
justinkambic: We don't have a formal decision made around this. Downstream I introduced `styled-components` and am trying to eliminate the inline styles. I made these changes before introducing that. Perhaps we could do a quick sync about how we want to handle custom styles in the project, as it's not something we've devoted attention to in the past.

src/components/Header/HeaderControls.tsx
```diff
@@ -0,0 +1,118 @@
...
+      </EuiFlexItem>
+      <EuiFlexItem grow={false}>
+        <EuiFlexGroup gutterSize="m">
+          <EuiFlexItem>
+            <TestButton disabled={disabled} onTest={onTest} />
+          </EuiFlexItem>
+          <EuiFlexItem
+            style={{
+              borderRight: euiTheme.border.thin,
+              paddingRight: 16,
```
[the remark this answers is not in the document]
justinkambic: I don't think so. As linked above React renders bare integer values to `16px;` in this case.

src/components/StepsMonitor.tsx
justinkambic: We can maintain this state wholly in the flyout now. Previously it was shared across the app but I think now it's only in the code flyout.

justinkambic: Done in https://github.com/elastic/synthetics-recorder/pull/107/commits/2b550099ffb26fdc1f312940e502c456eccd8f68.

justinkambic (commented):
I've implemented and answered a bunch of feedback, LMK your thoughts. Happy to discuss our approach to element-level styling as we haven't committed to any particular strategy.

src/common/shared.ts
```diff
@@ -108,3 +108,27 @@ export function updateAction(
...
+        step !== null &&
+        step.length > 0 &&
+        step[0].title &&
+        journeyStepNames.has(step[0].title)
+    ),
+    journey.type
+  );
+}
+
+export const SMALL_SCREEN_BREAKPOINT = 850;
```
vigneshshanmugam, line 134: nit: move constants together.

src/hooks/useSyntheticsTest.ts
```diff
@@ -0,0 +1,58 @@
...
+export function useSyntheticsTest(actions: ActionContext[][]) {
+  const [result, setResult] = useState<Result | undefined>(undefined);
+  const [codeBlocks, setCodeBlocks] = useState("");
+
+  const onTest = useCallback(
+    async function () {
+      /**
+       * For the time being we are only running tests as inline.
+       */
+      const code = await getCodeFromActions(actions, "inline");
```
vigneshshanmugam, line 40: Why is this hardcoded as inline? Did we remove the suite option?

src/components/Header/HeaderControls.tsx
```diff
@@ -0,0 +1,118 @@
...
+  hasActions: boolean;
+  setIsCodeFlyoutVisible: Setter<boolean>;
+}
+
+export function HeaderControls({
+  hasActions: disabled,
+  setIsCodeFlyoutVisible,
+}: IHeaderControls) {
+  const { euiTheme } = useEuiTheme();
+  const { isPaused, isRecording, toggleRecording } =
```
vigneshshanmugam, line 63: These types does not exist in recording context. Results in build error. 

```js
Property 'isPaused' does not exist on type 'IRecordingContext'.  TS2339

    61 | }: IHeaderControls) {
    62 |   const { euiTheme } = useEuiTheme();
  > 63 |   const { isPaused, isRecording, toggleRecording } =
       |           ^
    64 |     useContext(RecordingContext);
    65 |   const { onTest } = useContext(TestContext);
    66 |   return (
```

src/common/shared.ts
justinkambic: Fixed: https://github.com/elastic/synthetics-recorder/pull/107/commits/70234edb647b38dcdf140918bf54574bd1fd79d8.

src/hooks/useSyntheticsTest.ts
justinkambic: I seem to recall talking a while back about only running tests as `inline` - is this no longer accurate?

src/components/Header/HeaderControls.tsx
justinkambic: Happened because of a recent rebase, [fixed](https://github.com/elastic/synthetics-recorder/pull/107/commits/48708235ca4e00cd81bc620bf459962b8246e34b).

src/hooks/useSyntheticsTest.ts
vigneshshanmugam: For running yes, for extracting we did have both options right?

justinkambic: We should have both options. The updated design does need to account for this setting somewhere, that blocks merging this until we have an answer. I'll reach out to @liciavale about this.

As for this particular code is just for running the tests, so it should always be `inline` unless we support running tests as `suite`.

vigneshshanmugam: Sounds good, yeah we can do it separate ticket as well if you want. Lets not block this PR for this.

electron/execution.js
```diff
@@ -141,13 +141,17 @@ async function onTest(data) {
     switch (parsed.type) {
       case "journey/start": {
         const { journey } = parsed;
-        result.journeys[journey.name] = { status: "succeeded", steps: [] };
+        result.journey = {
+          status: "succeeded",
+          steps: [],
+          type: journey.name,
```
vigneshshanmugam, line 147: are we using this for suite vs inline?

vigneshshanmugam (approved):
LGTM, I didnt go through the code a lot as Lucas already did cover most of the points. 

Tested all the functionalities locally, Works great, Thanks @justinkambic 🎉

src/components/Header/HeaderControls.tsx
```diff
@@ -0,0 +1,118 @@
...
+      }}
+    >
+      <EuiFlexItem grow={false}>
+        <ControlButton
+          aria-label="Toggle script recording on/off"
+          color="primary"
+          iconType={isRecording ? "stop" : "play"}
+          onClick={toggleRecording}
+        >
+          {isRecording ? "Stop" : "Start recording"}
```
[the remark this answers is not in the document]
lucasfcosta: Agreed, it seems out of scope for this PR. Do we have anywhere we track these subjects and pending design decisions? Should we create an issue and then close it if it does not apply?

src/common/shared.ts
```diff
@@ -108,3 +114,49 @@ export function updateAction(
...
+  actions: ActionContext[][],
+  journeys: Journey | undefined,
+  type: string
+) {
+  if (!journeys) return "";
+  const journeyArr = combineResultJourneys(journeys);
+  const stepActions = journeyArr
+    .map(({ steps }) => {
+      for (const step of steps) {
+        return getStepActions(step, actions) ?? null;
```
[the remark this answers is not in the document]
lucasfcosta: Great work, looks much cleaner! Indeed, given we only deal with a single journey removing that pluralization, making it a single-journey and iterating through its steps is great and removes unnecessary looping (both syntax-wise in the variable names and semantically in the code).

src/components/Header/HeaderControls.tsx
lucasfcosta: Agreed, as in the previous item, how do you feel about having an issue for this or adding it to a Google Doc which we quickly review weekly as an "Agenda" for the recorder's roadmap?

src/components/Header/HeaderControls.tsx
```diff
@@ -0,0 +1,118 @@
...
+              borderRight: euiTheme.border.thin,
+              paddingRight: 16,
+            }}
+          >
+            <EuiButton
+              color="text"
+              iconType="editorCodeBlock"
+              onClick={async function () {
+                setIsCodeFlyoutVisible(true);
+              }}
```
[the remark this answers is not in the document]
lucasfcosta: Cool, that's NBD I just asked out of curiosity to see if we tend to enforce consistency. I don't really mind that much unless we're using lots of `this` 😄

src/components/TestButton.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const ariaLabel = disabled
+    ? "Record a step in order to run a test"
+    : "Perform a test run for the journey you have recorded";
+
+  const button = (
+    <EuiButton
+      aria-label={ariaLabel}
+      color="primary"
+      iconType="beaker"
+      isDisabled={disabled}
```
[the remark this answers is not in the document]
lucasfcosta: Re-tested and it was okay. Thank you!

lucasfcosta (approved):
LGTM!

The only thing that still happens on my machine is:

> It doesn't matter how much I expand the recorder window, it still shows this gap between the right side of the header and the window border.

But I'd be okay with merging this in and fixing it separately not to block this big batch of changes. We may want to take another look at this if we're going to release though, but I'll leave that up to you.

Also, for tracking outstanding discussion topics and issues, should we have issues for those or add them to an agenda? If you agree I'll add those to the backlog refinement meeting.

src/components/Header/HeaderControls.tsx
justinkambic: I'm find if we create an issue for this. We can link it in a roadmap agenda then. It won't be sustainable to use inline styles ad hoc, so we should formalize how we want to do this.

src/components/Header/HeaderControls.tsx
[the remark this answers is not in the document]
justinkambic: It's related to the design issue of deciding where/how to interact with this input. The latest design work we had for the URL input was placing it on a startup page, but @liciavale and I have some doubts about the UX of requiring users to back out to a start page anytime they wish to modify the URL. I created [an issue](https://github.com/elastic/synthetics-recorder/issues/123) to track this.

justinkambic:
> If you agree I'll add those to the backlog refinement meeting.

Sure thing, I did create an issue about the URL text input, but feel free to fill in others. I'm also happy to create the issues if you'd prefer.

> It doesn't matter how much I expand the recorder window, it still shows this gap between the right side of the header and the window border.

Interesting, this really is not happening for me on this branch.

PR #110, opened 2021-12-06
Add E2E test infrastructure

### LINKED ISSUE #26: Testing infrastructure for the recorder
To be able to release without any friction, we need to come. up with a small amount of Integration/E2E tests that covers the functionality.

### Overall

+ Launch recorder,
+ Start recording session by scripting it via Playwright
+ Able to generate Synthetics code correctly. 
+ Able to run the tests in two modes (inline, suite)
+ Save and export the script. 

Would like to have this as the first level as this brings in much more value than bunch of unit tests that tests the component functionality.

lucasfcosta:
> ⚠️ This PR supersedes https://github.com/elastic/synthetics-recorder/pull/104. The old PR with these tests has been automatically closed when I had to re-fork this repository given the fact it's gone public. Please review _this PR_, _not_ the old one.

# Summary

This PR adds the necessary infrastructure for running E2E tests.

Fixes #26.

It includes:
1. Running a packager to serve the recorder application itself
2. Obtaining an `electron` instance to interface with the application served within the electron executable
3. Connecting to the `chromium` browser spawned by the `electron`, so that we can perform actions and see whether they're captured

With the three items above, we can pretty much write any end-to-end test.

# Implementation details

In this section, I'll explain the 3 points from the summary: how we package things, how we open and control the recorder app in `electron` and how we control the `chromium` browser it spawns.

To make it easier for folks to understand how this testing infrastructure works, I've drawn a simple diagram:

e2e_1

1. Running the packager
    During our tests, we will spin-up the packager service as a separate process _before_ we even start `jest`. That packager runs on port `61337` by default. As the packager starts, we listen to what its process writes to `stdout` so that we know when the packager is ready to serve our JS application which will open within electron.
    **⚠️ IMPORTANT UPDATE:** The packager now only runs when developers use `npm run test:e2e:server`. Currently, we use `react:build` instead and have one-off runs using the built files. This change was so that we would be able not to worry about the packager during setup/teardown and so that users could still run the server and runner separately (through `npm run test:e2e:runner`) to iterate more quickly.
2. Running Electron
    The `electron` application, by default, will use the `electron` executable to open the `electron.js` script, which creates a new electron a main process, sets up IPC listeners to respond to actions from the renderer process, and connects to `localhost:3000`.
    For the app to connect to our packager server, all we need to do is to use the `TEST_PORT` environment variable to connect it to the port to which our test's packager is bound.
    As we start `electron`, we keep a reference to it via Playwright, which will automatically connect to it so that we have remote-control capabilities.
    We can use this instance to perform actions within our tests.
3. Controlling the remote browser
    This was the most challenging part of this PR.
    Because the `chromium` process spun up by `electron` is within the separate process which was initially spun up by node (see the diagram above), we can't simply use `jest` to intercept the `chromium.launch` call and keep references to the launched browsers. Instead, what I did was to pass a parameter to the `chromium` instance we open which will open up a remote port through which we can connect. Once connected, we can speak to that browser using the [`Chrome DevTools Protocol`](https://chromedevtools.github.io/devtools-protocol/).
    Now, just using the CDP would be quite annoying if we had to do it manually, so we use `playwright`'s `connectOverCDP` to actually get a `browser` instance (just like the `electron` one) on which we can call methods which will drive the remote chromium instance.

## Using a separate runner and server

As in Kibana, one can run the server with the packager separately from the test runner itself, so that they can iterate more quickly.

e2e_2

For that, you just need to run `npm run test:e2e:server` to kick-off the long-running server process, and run `npm run test:e2e:runner` to run your tests whenever you want.
    
# Next steps

## For this PR

This PR is still in `draft` state because I want to introduce the following minor changes:
* Making the CDP port configurable for new `chromium` instances
* Adding remote debugging capabilities to make it easier to test selectors (which is why the current assertions are disabled)

## For a further PR

* Adding a small web-application which will be served under `localhost` as we start the tests so that we can test against a controller environment
* Improve test/service APIs to make it more intuitive to write tests
* Setup CI with this infra


e2e/services/electron.ts
```diff
@@ -0,0 +1,100 @@
...
+          TEST_PORT: env.TEST_PORT,
+          PW_DEBUG: "console",
+          NODE_ENV: process.env.NODE_ENV,
+        },
+      });
+
+      return this.#instance;
+    } catch (e) {
+      // eslint-disable-next-line no-console
+      console.log(e);
```
vigneshshanmugam, line 50: Believe this should throw error instead?

e2e/services/electron.ts
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  async waitForPageToBeIdle(timeout = 45000) {
+    await this.#recordingBrowserPage.waitForLoadState("networkidle", {
+      timeout,
+    });
+  }
+
+  async navigateRecordingBrowser(url, waitUntilIdle = true, timeout = 4500) {
+    await this.#recordingBrowserPage.goto(url, { timeout });
```
vigneshshanmugam, line 91: wait Until here and waitIdle on the next line is so different in terms of load.

1. this.#recordingBrowserPage.goto(url, { timeout });. You can pass waitUntil: "networkidle" in this and it would wait until the URL loaded has done loading all the network requests.

2. The next line `waitForPageToBeIdle` does not have any context on what to wait on as the page state would have already been loaded and wont be based on any actions. 

I would squeeze them in and have it as part of line 91 `wait this.#recordingBrowserPage.goto(url, { waitUntil: "load"/"networkidle", timeout });

e2e/services/browser.ts
```diff
@@ -0,0 +1,92 @@
...
+    while (browserContext.pages().length === 0 || !browserContext.pages()[0]) {
+      await sleep(100);
+    }
+
+    return browserContext.pages()[0];
+  }
+
+  static async closeRemoteBrowser(connectParams?: ConnectRetryParams) {
+    const browser = await TestBrowserService.getRemoteBrowser(connectParams);
+    await browser.close();
```
vigneshshanmugam, line 90: would add an empty catch here, just in case.

e2e/services/electron.ts
```diff
@@ -0,0 +1,100 @@
...
+    await electronWindow.click("text=Start recording");
+    this.#recordingBrowserPage =
+      await TestBrowserService.getRemoteBrowserPage();
+    if (waitUntilIdle) {
+      await this.waitForPageToBeIdle();
+    }
+    return this.#recordingBrowserPage;
+  }
+
+  async waitForPageToBeIdle(timeout = 45000) {
```
vigneshshanmugam, line 84: As stated in the next comment, wait state depends a lot on previous action, so i would let the test script handle the condition.

e2e/setup/packager.js
```diff
@@ -0,0 +1,70 @@
...
+      env: {
+        PORT: env.TEST_PORT,
+        PATH: process.env.PATH,
+        BROWSER: "none",
+      },
+      shell: true,
+      detached: true,
+    });
+
+    ls.stdout.on("data", async data => {
```
vigneshshanmugam, line 40: nit: you could use async iterables to make this easier. Some prior art - https://github.com/elastic/synthetics-recorder/blob/0bf14641aad4ec49a06c70b555e81009d00f2def/electron/execution.js#L196-L201

e2e/setup/packager.js
```diff
@@ -0,0 +1,70 @@
...
+        await new Promise(resolve => killProcess.on("close", resolve));
+        resolve(packageFiles());
+      }
+
+      if (data.indexOf("You can now view") !== -1) {
+        resolve(ls);
+      }
+    });
+
+    ls.stderr.on("data", data => {
```
vigneshshanmugam, line 59: I would set encoding to `utf-8` for stdout and stderr stream

e2e/setup/packager.js
```diff
@@ -0,0 +1,70 @@
...
+        BROWSER: "none",
+      },
+      shell: true,
+      detached: true,
+    });
+
+    ls.stdout.on("data", async data => {
+      if (data.indexOf("Something is already running on port") !== -1) {
+        const killProcess = spawn(
+          `lsof -i 'tcp':${env.TEST_PORT} | grep 'LISTEN' | awk '{print $2}' | xargs kill -9`,
```
vigneshshanmugam, line 43: Can we warn and reject instead of closing the process?

e2e/setup/packager.js
```diff
@@ -0,0 +1,70 @@
...
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
+const { spawn } = require("child_process");
+const { env } = require("../services");
+
+const packageFiles = async () => {
+  return new Promise((resolve, reject) => {
+    const ls = spawn("npm", ["run", "react:start"], {
```
vigneshshanmugam, line 30: I thought about this for a while, do we want react:start or react:build here? Are we planning to test something during the middle of development? 

One issue is that we have to kill this process, instead if we relied on build, we could simply not care about the react-dev server during the teardown. WDYT?

package.json
```diff
@@ -17,8 +17,9 @@
     "lint": "eslint . --format eslint-formatter-multiple",
     "format": "prettier --write .",
     "lint:fix": "npm run lint -- --fix",
-    "test": "jest",
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "test:unit": "jest --config ./jest.unit.config.js ./src",
+    "test:e2e": "jest --config ./jest.e2e.config.js ./e2e/tests"
```
vigneshshanmugam, line 22: Need to add "test" and call test:unit and test:e2e there for the CI to run both for PR's.

```
 echo 'Error: no test specified'
```

e2e/setup/packager.js
lucasfcosta: That's definitely a good point to discuss, I've also thought about it as I was writing this PR.

The reason I kept it as `"react:start"` is because I planned on dettaching the packager from the overall routine so that we can separate the packager server and the runner (like Kibana does).

**However, thinking about this one more time I think you're right and it's definitely worth going for `react:build` straightaway for the `test:e2e` command.**

I have just updated this PR so that we have a `test:e2e:server` and `test:e2e:runner` command for using `react:start` and a separate `test:e2e` which simply uses `build`.

Please lmk what you think.

(Great comment btw! 🙌 )

e2e/services/browser.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    return browserContext;
+  }
+
+  static async getRemoteBrowserPage(connectParams?: ConnectRetryParams) {
+    const browserContext = await TestBrowserService.getRemoteBrowserContext(
+      connectParams
+    );
+
+    while (browserContext.pages().length === 0 || !browserContext.pages()[0]) {
```
justinkambic, line 84: From a resilience standpoint, any concern about throwing an error after some max interval here?

electron/electron.js
```diff
@@ -30,30 +30,38 @@ const debug = require("electron-debug");
...
       contextIsolation: false,
       nativeWindowOpen: true,
     },
   });
 
-  isDev
-    ? win.loadURL("http://localhost:3000")
-    : win.loadFile(join(BUILD_DIR, "index.html"));
+  if (isDev && !IS_TEST) {
+    win.loadURL("http://localhost:3000");
```
justinkambic, line 59: I'm just reflecting here, might be a nice future enhancement for this to not be hardcoded too. Out of scope for this patch of course.

src/common/shared.test.ts
```diff
@@ -24,10 +24,11 @@ THE SOFTWARE.
 
 /* eslint-disable import/first */
 window.require = require;
-
 import { updateAction } from "./shared";
 import { ActionContext } from "./types";
 
+window.require = require;
```
justinkambic, line 30: Interesting - when first adding unit tests this only worked with the override before these imports.

We might want to looking into adding this globally since we're going to need it in many test files in the future.

justinkambic (commented):
Great work here! Had a few minor comments, but I pulled this and it runs well for me.

e2e/services/electron.ts
vigneshshanmugam: Nit: console.error.

src/common/shared.test.ts
vigneshshanmugam: duplicated.

e2e/setup/builder.js
```diff
@@ -0,0 +1,52 @@
...
+        PATH: process.env.PATH,
+        BROWSER: "none",
+      },
+      shell: true,
+      detached: true,
+    });
+
+    ls.on("close", resolve);
+
+    ls.stderr.setEncoding("utf8");
```
vigneshshanmugam, line 40: Its good to also log the stdout as we are waiting for the react build to happen and it felt like there was nothing happening in the terminal.

vigneshshanmugam (commented):
LGTM, Added few comments and seems like child process is not exiting correctly which triggers this error when runnin E2E. 

```
Jest did not exit one second after the test run has completed.

This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
```

vigneshshanmugam (approved):
LGTM pending test failure.

justinkambic (approved):
LGTM!

.ci/Jenkinsfile
```diff
@@ -24,59 +24,64 @@ pipeline {
...
-    stage('Install'){
-      steps {
-        withGithubNotify(context: "Install") {
-          withNodeJSEnv() {
-            dir("${BASE_DIR}") {
-              sh(label: 'npm install', script: 'npm install')
-            }
+    stage("Prepare environment") {
+      def environment = docker.build("test-image", "${BASE_DIR}/e2e/Dockerfile")
+      environment.inside {
```
v1v, line 29: Can you confirm this is possible to use? As far as I know, the declarative pipeline syntax is quite restrictive when using groovy specifics, and for such it requires to use the `scripts` closure, normally within the `steps` closure.

lucasfcosta: Yeah, this definitely didn't work. I ended up trying to use the `dockerfile` and `docker` agents to configure build steps. No luck so far, but I have a few more things to attempt. Thanks for your input 🙌

e2e/Dockerfile.jenkins
```diff
@@ -0,0 +1,54 @@
...
+RUN usermod -s /bin/bash jenkins
+
+# Copying repo directory
+ADD --chown=jenkins $BASE_DIR $DOCKER_BASE_DIR
+
+USER jenkins
+
+# Installing NVM
+ENV NVM_DIR /home/jenkins/.nvm
+RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.0/install.sh | bash
```
lucasfcosta, line 44: I'd appreciate some input on whether you think we should change this.

Given it runs with the Jenkins user's privileges I don't think it'll be a problem to pipe install into bash, but I'm also not a specialist in these types of things so I'd appreciate further guidance here on whether we need to change this in any way.

CC @justinkambic @vigneshshanmugam @v1v

vigneshshanmugam: I am fine with this approach. Btw, Is it possible to build this image and push it to observability docker registry which has xvfb support instead of doing this on each PR? /cc @v1v Thoughts?

vigneshshanmugam (approved):
Great work @lucasfcosta and the docs are 🚀  , Lets get this in.

justinkambic (approved):
LGTM!

e2e/Dockerfile.jenkins
v1v: > which has xvfb support instead of doing this on each PR? /cc @v1v Thoughts?

It's possible to make use of the `packer_cache.sh` script, this particular script is provided by the infra-automation team to cache the required dependencies in the CI Workers.

For such, it's required some changes:
1) Ask/Open a PR in infra to support this project for this particular caching feature
2) Create the `.ci/packer_cache.sh` script similar to -> https://github.com/elastic/apm-agent-rum-js/blob/master/.ci/packer_cache.sh
3) Support in the CI to build the docker image with the cached docker images, when possible. NOTE: if there are changes in the dockerfile then it will be required to be built!
4) Agree with the docker tag in the `packer_cache.sh` and the Jenkinsfile.

.ci/Jenkinsfile
```diff
@@ -27,9 +29,9 @@ pipeline {
     stage('Checkout') {
       steps {
         pipelineManager([ cancelPreviousRunningBuilds: [ when: 'PR' ] ])
-        deleteDir()
-        gitCheckout(basedir: "${BASE_DIR}", githubNotifyFirstTimeContributor: true)
-        stash allowEmpty: true, name: 'source', useDefaultExcludes: false
+          deleteDir()
+          gitCheckout(basedir: "${BASE_DIR}", githubNotifyFirstTimeContributor: true)
+          stash allowEmpty: true, name: 'source', useDefaultExcludes: false
```
v1v, line 34: NIT: unrequired :)

```suggestion
        deleteDir()
        gitCheckout(basedir: "${BASE_DIR}", githubNotifyFirstTimeContributor: true)
        stash allowEmpty: true, name: 'source', useDefaultExcludes: false
```

.ci/Jenkinsfile
```diff
@@ -56,16 +58,31 @@ pipeline {
...
-              sh(label: 'npm test', script: 'npm test')
+    stage('Docker test setup') {
+      agent {
+        dockerfile {
+          reuseNode true
+            dir "$BASE_DIR"
+            filename './e2e/Dockerfile.jenkins'
+            additionalBuildArgs  "--build-arg BASE_DIR=\".\" --build-arg DOCKER_BASE_DIR=\"${DOCKER_BASE_DIR}\""
+        }
+      }
```
v1v, line 73: I'm not sure if the `packer_cache.sh` approach works well with this specific declarative approach.

I wonder if this section could be embedded in a shell step itself:

1) Create entrypoint:
```bash
#!/bin/bash
Xvfb ${DISPLAY} -screen 0 1024x768x16 &
npm test
                
source $NVM_DIR/nvm.sh
cd $DOCKER_BASE_DIR
nvm use
npm test
```

2) Create script in the path `.ci/scripts/run-e2e.sh` (or somewhere else)
```bash
docker build --build-arg BASE_DIR=\".\" --build-arg DOCKER_BASE_DIR=\"${DOCKER_BASE_DIR}\" -f ./e2e/Dockerfile.jenkins -t synthetics-recorder:latest .
docker run --rm -ti synthetics-recorder:latest
```

3) Edit pipeline to run

```
stage('Docker test setup') {
  steps {
      withGithubNotify(context: "Test") {
        sh(label: 'run e2e', script: '.ci/scripts/run-e2e.sh')
      }
  }
}
```

The above approach will avoid any dependency with the CI, since what you can run locally in one script can be executed in the CI with one call.

PR #114, opened 2021-12-06
chore: Add `nodemon`

justinkambic:
Today when we make changes to our `electron` code, we need to restart the process to see them reflected. This can be frustrating when debugging or adding new functionality.

These changes introduce `nodemon` to manage the restart for us manually when we're targeting the electron files. It shouldn't need to watch anything else since react scripts will automatically reload the frontend files for us.

### DIFF
--- a/nodemon.json
+++ b/nodemon.json
@@ -1,0 +1,3 @@
+{
+  "watch": ["electron/"]
+}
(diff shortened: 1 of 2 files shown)


src/common/shared.test.ts
```diff
@@ -28,6 +28,8 @@ window.require = require;
 import { updateAction } from "./shared";
 import { ActionContext } from "./types";
 
+window.require = require;
```
vigneshshanmugam, line 31: this is duplicated again. Why?

justinkambic: Rebase error 😭

PR #115, opened 2021-12-06
Add local WebApp to the E2E testing infrastructure

### LINKED ISSUE #112: [Testing Infrastructure] There's a local web application available in our E2E tests for us to record journeys against
To facilitate recording tests under a controlled environment during E2E tests, our E2E test setup should spin-up a web-application instance. This web application should be, preferably, the e-commerce application we use to demo our synthetics journeys.

Using such a web application allows us not to depend on third-parties for our tests to pass.

lucasfcosta:
⚠️ This PR has been made on top of #110, so please review and merge that PR first.

# Summary

This PR adds a local WebApp to the E2E testing infrastructure so that we can write tests without relying on third party websites which may change.

Closes #112.

## How this works

In addition to starting the packager, we also use the native `http` module and `next` to start a server in the directory which contains a demo app.

This demo app is [the same `e-commerce-demo` we use in the `synthetics-demo` repository](https://github.com/vigneshshanmugam/synthetics-ecommerce-demo).

Please notice that the dependencies for that nested project are installed in our `postinstall` script, which runs automatically after `npm install`. We also build the `demo-app` during the `postinstall` script.

Once set-up, the promise returned by the function which starts the next server will assign the server to the global namespace and resolve. Then, during teardown, we access the global namespace and stop the `http` server.

e2e_3

Given this application will be running all throughout our tests, we can simply use an address pointing to `localhost` to access it.

Please notice that I also reorganised a few files as I was doing this change.

## Setting up the nested project

I'm using git subtrees to manage the separate e-commerce app, which is what I think @devcorpio was referring to in our tech sync earlier today.

It is quite straightforward and doesn't add any maintenance burden unless you're rebasing the two unrelated trees from which the merge commit from the subtree comes from.

Given we won't have to rebase from before the subtree merge point once this PR is merged, I think subtrees are the way to go. **I've also looked into alternatives but all other products seemed unnecessarily complicated, or addressed somewhat different problems (as Lerna does)**.


e2e/demo_app/LICENSE
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
```
vigneshshanmugam, line 1: nit: don't think we need this.

e2e/demo_app/index.html
```diff
@@ -0,0 +1,69 @@
...
+      <p>Written by <a href="http://evanyou.me">Evan You</a></p>
+      <p>Part of <a href="http://todomvc.com">TodoMVC</a></p>
+    </footer>
+
+    <script>
+    // for testing
+    if (navigator.userAgent.indexOf('PhantomJS') > -1) localStorage.clear()
+    </script>
+    <!-- Delete ".min" for console warnings in development -->
+    <script src="vue.min.js"></script>
```
vigneshshanmugam, line 66: WDYT? can we fetch it from `https://unpkg.com/browse/vue@2.6.14/dist/vue.min.js` than having this local?

e2e/setup/demo_app.js
```diff
@@ -0,0 +1,48 @@
...
+const koaStatic = require("koa-static");
+
+const { env } = require("../services");
+
+const DEMO_APP_DIR = path.join(__dirname, "../demo_app");
+
+const startDemoApp = () => {
+  return new Promise(resolve => {
+    const koaInstance = new Koa();
+    koaInstance.use(koaStatic(DEMO_APP_DIR));
```
vigneshshanmugam, line 37: Nothing aganist using Koa and its sub modules. But keeping the complexity and scope simpler, we can use a simpler static server than bringing in a koa framework and its middlewares. https://github.com/elastic/synthetics-demo/blob/6fc9ae3388b1054e6ff795c500544c496e4782ee/todos/synthetics-tests/hooks.journey.ts#L10-L12

vigneshshanmugam (commented):
Reviewed only the last commit, Looks good to me. Added small suggestions, but not blockers.

e2e/setup/demo_app.js
lucasfcosta: Much better, thank you! I was looking exactly for this before I started using Koa, but a Google search only pointed to even bigger software, like Express and similar things. Thank you, I'm all for _less_ stuff :D

demo-app/yarn.lock
```diff
@@ -0,0 +1,4063 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
vigneshshanmugam: please remove this file, we are using NPM in the project and having yarn lock doesn't make sense.

PR #119, opened 2021-12-09
Use npm ci instead of npm install for CI pipeline

lucasfcosta:
## Summary

This PR changes the command we use to install dependencies on CI from `npm install` to `npm ci`.

## Implementation details

When using `npm install` you get the newest "non-breaking" versions of packages (according to the authors which versioned it). That means you may get packages with changes that cause the build to behave differently, and that could even cause the build to fail, as one cannot rely on third parties to follow `semver` strictly.

The `npm ci` command, on the other hand, uses the `package-lock.json` file to get the exact same dependencies that you had in your machine, down to the patch version. That ensures our builds will be repeatable.

It also makes the build slightly faster, although not by much, as we don't have to resolve newest versions. We're essentially doing a frozen build.

[I've previously written more about this subject in this post.](https://lucasfcosta.com/2020/10/17/lockfile-guide.html).

Docs for `npm ci` are [here](https://docs.npmjs.com/cli/v8/commands/npm-ci).

You can see that CI now runs `npm ci` instead of `npm install`

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -37,7 +37,7 @@ pipeline {
         withGithubNotify(context: "Install") {
           withNodeJSEnv() {
             dir("${BASE_DIR}") {
-              sh(label: 'npm install', script: 'npm install')
+              sh(label: 'npm ci', script: 'npm ci')
             }
           }
         }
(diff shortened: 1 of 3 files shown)


justinkambic (approved):
LGTM

PR #130, opened 2021-12-21
Ignore `demo-app/package-lock.json`

justinkambic:
## Summary

Purpose here is to ignore a `package-lock.json` if it's created on the demo app during an install run.

## Implementation details

Also, this added an empty line at the bottom of the `package.json`, I'm not sure if this will differ based on the dev platform.

Run `npm i` or `npm ci` and see if a `package-lock.json` shows up in the demo app directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ local-browsers
 .synthetics
 react-app-env.d.ts
 eslint-junit.xml
+demo-app/package-lock.json
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,3 @@ local-browsers
 .synthetics
 react-app-env.d.ts
 eslint-junit.xml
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -7,3 +7,4 @@ local-browsers
 .synthetics
 react-app-env.d.ts
 eslint-junit.xml
+demo-app/package-lock.json
```
vigneshshanmugam, line 10: It should be part of .npmrc file that ignores lock files.

justinkambic: I'll add it but it seems only to work when running an `npm install` from the `demo-app` directory, so we're going to want to keep ignoring it if we aren't going to commit it.

vigneshshanmugam (approved):

PR #133, opened 2022-01-04
Update dependencies

lucasfcosta:
## Summary

This PR updates a few dependencies which were recommended to us by our automations.

It includes updates to:
* EUI (to solve problems for for the downstream dependencies)
* `trim`
* `react-scripts`

Furthermore, it also adds a missing dependency (`styled-jsx`) to our `demo-app`.

⚠️ Please notice that we couldn't update `ansi-html` because there's no fix available upstream. It shouldn't be a problem though given that `ansi-html` seems to only be used by `webpack-dev`.

⚠️ It seems like there were also changes to Jenkins which change its UID when it logs into the docker container, therefore I've now updated the Jenkinsfile not to depend on the Jenkins user anymore so that builds are more reliable.

## Implementation details

There are two important implementation details to notice about this PR:

1. I've had to use [`npm-force-resolutions`](https://www.npmjs.com/package/npm-force-resolutions) to force `trim` to resolve to necessary version.
    I had to use the `npm-force-resolutions` package because `trim` was used by a dependency of EUI, which wasn't updated upstream.
2. The UI changed slightly with the update to EUI.
    I did try to make it look as much like the previous version as possible, but a few changes weren't parameterised, so I assume that's because the design team wants components to look that way across all products.
    ⚠️ Please also notice that the theme we were using is now a `legacy` theme. You can see this change in EUI's repo [here](https://github.com/elastic/eui/pull/5121/files#diff-0629aaf7668deed639f57d6a41308b1ebdc0f1bbc974108bc4dfad77e1b49e2aR55).

1. Run an installation and all tests locally.
2. Use the recorder to record a couple of scripts and ensure it still works.
3. Try to export code (using the fly-out too).
4. Navigate around and make sure you're happy with its overall look and feel.
5. Confirm you're happy with the visual changes due to the EUI update.
    **Before**
    Screenshot 2022-01-04 at 11 31 16
    **After**
   Screenshot 2022-01-04 at 13 41 00


justinkambic (commented):
There's some weirdness going on with the styling of the corners of the step containers:

image

If you want I am happy to pair and see if we can track this down.

EDIT: this particular component is going to go away before the next release, so I am not overly concerned with this style issue. I have a body update I am planning to publish soon, but I'll wait until we get this merged so I can incorporate the updated UI styles into it.

justinkambic (approved):
LGTM! Thanks for taking on the chore and getting rid of those dep warnings.

PR #138, opened 2022-01-20
chore: Rely on synthetics `ActionInContext` and `Action` types

### LINKED ISSUE #85: chore: Depend on action types from Synthetics
Today, we define our own [types](https://github.com/elastic/synthetics-recorder/blob/main/src/common/types.ts#L12) for actions.

The [Synthetics](https://github.com/elastic/synthetics) project also defines very similar action types, that are slightly different from our own. It would be nice if Synthetics exported these types so the recorder could depend on them. This issue will require upstream modification of Synthetics to export the necessary types.

justinkambic:
## Summary

Resolves #85.

Depends on https://github.com/elastic/synthetics/pull/443. Cannot be merged until the next Synthetics release after this PR is merged.

Removes the `ActionContext` and `Action` types from the recorder and imports the corresponding types from `@elastic/synthetics`.

There is some mis-match between the types in the recorder and those in Synthetics, so it will require more upstream modification before we can resolve this.

## Implementation details

Simply replaces the types we've defined in this project with the ones from the Synthetics dependency.

If typescript builds without complaining and CI passes there should be no further testing action needed.


src/helpers/generator.ts
```diff
@@ -22,9 +22,10 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 */
 
-import { Action, ActionContext } from "../common/types";
+import type { ActionInContext } from "@elastic/synthetics";
+import { Action } from "../common/types";
```
kyungeunni, line 26: a minor comment(probably it's just my preference so feel free to ignore).
 I wonder if we can export these two from the same source. `Action` type is derived from `ActionInContext` and seeing them imported from different sources seems like they are unrelated.

kyungeunni (approved):

src/helpers/generator.ts
justinkambic: Yes we can modify synthetics to also export the `Action` type. If we do this we'll want to hold off on merging this until the next release of the synthetics library, which is fine. This change is low-priority.

vigneshshanmugam (approved):

lucasfcosta (approved):
LGTM, seems pretty straightforward.

PR #139, opened 2022-01-20
feat: Disallow test run during recording session

### LINKED ISSUE #127: [Enhancement] Users should not be able to replay steps during test recording
### Feature summary

Currently, users can click to replay test's steps _during_ the recording. We should only enable the `replay` feature once users have stopped the recording (or paused it, if we decide to allow that).

Even though this feature works fine, it's confusing that users could replay steps as they record journeys.

Disabling this feature would make the expected flow clearer.

### Linked resources

_No response_

### Additional information

We still need to discuss whether we'd like to implement this enhancement. We also need to discuss the possibility of only being able to `stop` a recording (removing the `pause`), as that may influence the desired behaviour for this enhancement.

justinkambic:
## Summary

Resolves #127.

Makes it so the user can't have the recorder run their script until they've finished recording.

## Implementation details

We've decided that it's a better experience if the user doesn't attempt to run their test until after they terminate their recording session.

1. Run the recorder
2. Start a recording session
3. Observe that the replay steps button is disabled
4. Close your Chromium recording session
5. Observe that you can now replay the steps you recorded

### DIFF
--- a/src/components/TestButton.tsx
+++ b/src/components/TestButton.tsx
@@ -36,7 +36,7 @@ export function TestButton({ isDisabled, onTest }: Props) {
     <ControlButton
       aria-label={
         isDisabled
-          ? "Record a step in order to run a test"
+          ? "You cannot test your recorded tests until you have finished a recording session"
           : "Perform a test run for the journey you have recorded"
       }
       color="primary"
--- a/src/components/TestButton.tsx
+++ b/src/components/TestButton.tsx
@@ -36,7 +36,7 @@ export function TestButton({ isDisabled, onTest }: Props) {
     <ControlButton
       aria-label={
         isDisabled
-          ? "You cannot test your recorded tests until you have finished a recording session"
+          ? "You cannot execute your recorded tests until you have finished a recording session"
           : "Perform a test run for the journey you have recorded"
       }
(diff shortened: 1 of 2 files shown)


lucasfcosta (approved):
LGTM, although I think we should eventually add a test for this one. What do you think? 🙂

src/components/TestButton.tsx
```diff
@@ -36,7 +36,7 @@ export function TestButton({ isDisabled, onTest }: Props) {
     <ControlButton
       aria-label={
         isDisabled
-          ? "Record a step in order to run a test"
+          ? "You cannot test your recorded tests until you have finished a recording session"
```
lucasfcosta, line 39: In terms of copy, I think `test your recorded tests` sounds a bit weird. I'd probably use "_execute_ your recorded tests" or "_run_ your recorded tests".

lucasfcosta (approved):
I just had a comment about the copy, but the code itself LGTM.

src/components/TestButton.tsx
justinkambic: I'll push this update, well-spotted. Result of writing copy while focused on something else.

PR #144, opened 2022-02-01
chore: Remove direct `ipc` imports, centralize import with `CommunicationContext`

### LINKED ISSUE #142: [Other] Single point of entry for `ipc`
### Task summary

In numerous places on the front end code, we are importing the `ipc` client. This makes it tricky to write unit tests and is just kind of a messy pattern. We should instead import the `ipc` in one place, preferably `App.tsx`, and then make it available to the rest of the application from a centralized point, likely via a context named something like `IPCContext` or `CommunicationContext`, etc.

More details below.

### Linked resources

_No response_

### Additional information

Example implementation found in today's version of `shared.ts`:

```typescript
import React from "react";
import type {
  ActionContext,
  Journey,
  JourneyType,
  Setter,
  Steps,
} from "./types";

const { ipcRenderer: ipc } = window.require("electron-better-ipc");

// ...

export function performSelectorLookup(
  onSelectorChange: Setter<string | undefined>
) {
  return async () => {
    const selector = await ipc.callMain("set-mode", "inspecting");
    if (selector) {
      onSelectorChange(selector);
      await ipc.callMain("set-mode", "recording");
    }
  };
}
```

We should instead strive for our files to look like this, which will make mocking much simpler and result in generally cleaner files and less potential setup/teardown requirements:

```typescript
import React from "react";
import type {
  ActionContext,
  Journey,
  JourneyType,
  Setter,
  Steps,
} from "./types";

// ...

export function performSelectorLookup(
  ipc: RendererProcessIpc, // <~~ the function defines ipc as a dep, rather than doing file-level import
  onSelectorChange: Setter<string | undefined>
) {
  return async () => {
    const selector = await ipc.callMain("set-mode", "inspecting");
    if (selector) {
      onSelectorChange(selector);
      await ipc.callMain("set-mode", "recording");
    }
  };
}
```

justinkambic:
## Summary

Resolves #142

## Implementation details

Per the linked issue, we have numerous imports of `electron-better-ipc` via `window.require`, which works fine in production but makes the code excruciating to test. Rather than introduce strange testing patterns throughout the codebase, I propose we use this `CommunicationContext`, which uses `window.require` as a single point of entry for `ipc`. 

We can then share this reference throughout the application, and modify any functions that were performing a direct import to use dependency injection instead. This will allow us to supply mocks very easily for any component or helper function that previously performed the import internally.

Run the recorder, record some steps, view exported code, run test and see the results. All of these activities will ensure that the IPC is still working ok.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -22,7 +22,7 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 */
 
-import React from "react";
+import React, { useContext } from "react";
 import { useEffect, useRef, useState } from "react";
 import { EuiFlexGroup, EuiFlexItem, EuiPageBody } from "@elastic/eui";
 import "./App.css";
@@ -34,6 +34,7 @@ import { AssertionDrawer } from "./components/AssertionDrawer";
 import { Title } from "./components/Header/Title";
 import { HeaderControls } from "./components/Header/HeaderControls";
 import { AssertionContext } from "./contexts/AssertionContext";
+import { CommunicationContext } from "./contexts/CommunicationContext";
 import { RecordingContext } from "./contexts/RecordingContext";
 import { StepsContext } from "./contexts/StepsContext";
 import { TestContext } from "./contexts/TestContext";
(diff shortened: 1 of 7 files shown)


vigneshshanmugam (approved):

kyungeunni (approved):
LGTM! :shipit:

PR #145, opened 2022-02-01
chore: add watch script for unit test DX, ignore `coverage` dir

justinkambic:
## Summary

Adds a `--watch` option for the npm test script with coverage flag set to aid with the unit test developer experience.

## Implementation details

Run `npm run test:unit:watch` to start `jest` in watch mode. `--coverage` flag is also set to generate coverage report for checking code paths within functions.

No user impact.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "prepare": "husky install",
     "test": "npm run test:unit && npm run test:e2e",
     "test:unit": "jest --config ./jest.unit.config.js ./src",
+    "test:unit:watch": "jest --config ./jest.unit.config.js ./src --coverage --watch",
     "test:e2e": "jest --config ./jest.e2e.config.js ./e2e/tests",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -22,6 +22,7 @@
     "prepare": "husky install",
     "test": "npm run test:unit && npm run test:e2e",
     "test:unit": "jest --config ./jest.unit.config.js ./src",
+    "test:unit:watch": "jest --config ./jest.unit.config.js ./src --coverage --watch",
```
kyungeunni, line 25: How about referring to existing script assuming we want to add run `test:unit` with additional options:
```suggestion
    "test:unit:watch": "npm run test:unit -- --coverage --watch",
```

kyungeunni (approved):

package.json
justinkambic: Great suggestion, will modify.

kyungeunni (approved):

PR #146, opened 2022-02-02
chore: Implement RTL helper to reduce test boilerplate

justinkambic:
## Summary

Today, we have many react components which rely on contexts for sharing state/functionality across the app without duplication or requiring re-nesting properties far down the component tree. This works very well for quickly sharing/implementing functionality. Unfortunately, reliance on `Context` comes at a cost in several areas. One of these areas is testing; because many components could rely on a context, that context must be implemented for each test. Because we haven't tested many components yet, we haven't really dealt with this cost.

As we start to get closer to a v1 level of quality, we're going to need to implement unit tests for components. In preparation for this, I've taken a stab at implementing some helper functions for our front-end to make it much quicker to write unit tests for React components.

## Implementation details

This code essentially introduces some middleware for the RTL `render` function. It overlays our contexts with some basic defaults defined that can be easily overridden. The RTL interaction on the caller side will remain unchanged.

You can modify a test to try this out if you wish, there are no user-facing effects.

### DIFF
--- a/src/helpers/test/RenderContexts.tsx
+++ b/src/helpers/test/RenderContexts.tsx
@@ -1,0 +1,62 @@
+/*
+MIT License
+
+Copyright (c) 2021-present, Elastic NV
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
+import React from "react";
+
+export interface ContextRenderer<ContextValue> {
+  defaults: ContextValue;
+  Context: React.Context<ContextValue>;
+  overrides?: Partial<ContextValue>;
+}
+
+/**
+ * Takes a list of React Contexts and values and recursively renders them overtop of a given component.
+ * @returns Rendered contexts wrapping child component.
+ */
+export const RenderContexts = ({
+  contexts,
+  children,
+}: {
+  /* The below line circumvents TS screaming about the different context types not being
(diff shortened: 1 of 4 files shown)


src/helpers/test/RenderContexts.tsx
```diff
@@ -0,0 +1,62 @@
...
+export const RenderContexts = ({
+  contexts,
+  children,
+}: {
+  /* The below line circumvents TS screaming about the different context types not being
+   * compatible with a union of all types, i.e. "IRecordingContext is incompatible with IRecordingContext | IStepsContext".
+   *
+   * Given the purpose of this function is to recursively render a generic list of contexts, it's not
+   * imperative to know or care what the type of the context is.
+   */
```
lucasfcosta, line 46: Thanks for the detailed explanation, really useful!

src/helpers/test/RenderContexts.tsx
```diff
@@ -0,0 +1,62 @@
...
+}) => {
+  const context = contexts.shift();
+  if (typeof context === "undefined") {
+    return <>{children}</>;
+  }
+  const { defaults, Context, overrides } = context;
+  const value = overrides ? { ...defaults, ...overrides } : defaults;
+  return (
+    <Context.Provider value={value}>
+      <RenderContexts contexts={contexts}>{children}</RenderContexts>
```
lucasfcosta, line 59: Really elegant recursive implementation, I love it.

src/helpers/test/defaults.ts
```diff
@@ -0,0 +1,43 @@
...
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
+import { RecordingStatus } from "../../common/types";
+import { IRecordingContext } from "../../contexts/RecordingContext";
+import { IStepsContext } from "../../contexts/StepsContext";
+import { IUrlContext } from "../../contexts/UrlContext";
+
+export const RECORDING_CONTEXT_DEFAULTS: IRecordingContext = {
```
lucasfcosta, line 30: I wonder whether we should make these functions which return new objects every time to avoid mutation and the history of previous mocks interfering with other test's results.

src/helpers/test/render.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { IStepsContext, StepsContext } from "../../contexts/StepsContext";
+import { IUrlContext, UrlContext } from "../../contexts/UrlContext";
+import {
+  RECORDING_CONTEXT_DEFAULTS,
+  URL_CONTEXT_DEFAULTS,
+  STEPS_CONTEXT_DEFAULTS,
+} from "./defaults";
+import { RenderContexts } from "./RenderContexts";
+
+export function render<ComponentType>(
```
lucasfcosta, line 44: Please have a look at my comment with regards to creating new objects on every render to avoid persistent state from one test to another.

lucasfcosta (changes requested):
What do you think about my comment with regards to creating new objects with default values every time we have to render? My concern is that one test will interfere with the other if the context gets mutated.

src/helpers/test/defaults.ts
justinkambic: Sounds like a great idea. I'll work on this modification now.

justinkambic: https://github.com/elastic/synthetics-recorder/pull/146/commits/3fb3e9399b0bffd8d48e923f04684a90a09ec273

LMK if anything looks wrong with this.

src/helpers/test/render.tsx
justinkambic: I'm +1 to this.

src/helpers/test/RenderContexts.tsx
justinkambic: Ah thanks, I was concerned it was overly clever, but my hope is that it will work seamlessly enough that no one will ever have to look at it.

justinkambic:
Thanks for the great feedback! ❤️

PR #152, opened 2022-02-11
feat: add result count e2e

justinkambic:
## Summary

Adds another e2e test, which verifies that the result count shows up as expected.

## Implementation details

We had some flaky/infra issues adding a second test. We've removed the external website dependency from the `navigation.test.ts` suite, and added a basic web server to check against navigations.

Run `npm run test:e2e` repeatedly and verify that the tests are both passing.


lucasfcosta (approved):
Excellent work! Thanks for taking care of this. It was a fun one to debug 😄

Issue #153, opened 2022-02-14
[Bug] Recorder is not recording all clicks


lucasfcosta:
### Bug summary

As shown in the GIF below, I've seen that sometimes the recorder will not properly record clicks on certain links.

![not_recording](https://user-images.githubusercontent.com/6868147/153912007-7962fab0-dff7-43a7-8972-b99faaf8c827.gif)

I've seen this behaviour is really easy to reproduce when clicking on the navbar icons in `bbc.co.uk`, but not easy to reproduce on other simpler websites.

It is, however, a bug that's common enough to warrant some investigation, I think.

### Recorder Version

Happening on the top of `main` (`812f7bf713889f1364b3a88cc0308e9c0ba8bb30`) and `0.0.1-alpha.3`

### Steps to reproduce

1. Enter `bbc.co.uk` in the URL field
2. Start recording
3. Click a common link in the middle of the page and you'll see it gets recorded
4. Click a link in the page's navbar and you'll see that most of the time it does not get recorded

### Expected behavior

All interactions with the web page should be recorded.

lucasfcosta:
I've just re-tested this after updating my MacOS and this issue doesn't seem to occur anymore.

I believe it might have been related to my machine's utilisation being too high with me running a quite a few instances of the Stack + some load generators.

To confirm, I've run the same containers (which cause my machine to go significantly slower), and tried to perform the same test. Once I did that, the problem did reoccur a few times.

@justinkambic @paulb-elastic I'd suggest that we have at least one more person to try to reproduce this so that we can confirm this doesn't happen on someone else's machine under normal circumstances. Then, I'd say we may need to discuss whether we want to make the recorder more resilient under these performance circumstances (and whether we can do that ourselves or if we need to do it upstream on Playwright).

justinkambic:
FWIW I did try this out while running a Windows and Ubuntu VM, with all three operating systems building/releasing the recorder at the same time. I think something like `63GB` of memory was consumed, and I was able to run the recorder and have it follow my journey. Not a scientific test by any means, but it was a decent best effort. I didn't notice any problems.

lucasfcosta:
I'll close this for now given the test @justinkambic has just described and that it seemed to be related exclusively to my machine running way too many Docker containers and having its HW highly stress tested.

Please feel free to reopen if you feel like we should actually do any more work or profiling on this.

PR #157, opened 2022-02-14
chore: Bump deps

justinkambic:
## Summary

Updates a number of project dependencies.

## Implementation details

This updates a number of our project dependencies to new versions. Most importantly, EUI is bumped to its latest major version, which has official support for the version of React we're using.

These changes should be relatively transparent. If all tests are passing and a cursory smoke test of the various user flows appears to work there should be no further cause for concern.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1705,17 +1705,17 @@
       }
     },
     "@elastic/eui": {
-      "version": "43.1.1",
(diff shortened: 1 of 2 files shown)


lucasfcosta (approved):
LGTM. I've also tested this PR E2E and it works fine. I've also updated #153 during that test as I think my machine's performance may have been what caused it.

PR #159, opened 2022-02-16
feat: Show all action titles in step containers for test result flyout

### LINKED ISSUE #131: Replay/Test steps flow
As a user of the script recorder

- [x]  I want to test the script I created
- [x] I want to see the steps that are successful/skipped/failed
- [x] When a step is failing I want a summary that helps me to identify what went wrong 

In order to identify any underlying issues/errors, I will need to fix them.

<details>
<summary>Deliverables</summary>

test

- [Figma file](https://www.figma.com/file/6xfG7uIpgc84d62JnXhOMV/Script-recorder?node-id=696%3A119428) 

- [Prototype ](https://www.figma.com/proto/6xfG7uIpgc84d62JnXhOMV/Script-recorder?node-id=848%3A128541&scaling=scale-down&page-id=696%3A119428&starting-point-node-id=696%3A119429)
</details>

justinkambic:
## Summary

Resolves #131.

As pictured below, shows all the actions in the result panels:

image

### Things to note

I have moved the duration value to the top of the step panel, because otherwise the duration element is appended to all of the actions and is misleading. Today, we only have resolution to the step level in terms of duration.

I am working on some [upstream changes](https://github.com/elastic/synthetics/pull/457) to Synthetics that should allow us to get action-level performance. At that point, we should be able to supply duration value for each item in this list, along with a cumulative duration value displayed in the heading.

Because we don't have action-level resolution, this means that all actions within a step currently show as the same status. If one action fails, all the actions in the step need to be marked as failure, because we currently do not know which one it was. Once the above changes are merged, we should be able to fix this.

## Implementation details

The backend electron code has been updated to find and supply the titles for all the actions contained in a given step. Given the constraints we have imposed on how step names are interpolated by the backend, we should always be able to parse out the original step data and retrieve the action titles for it, based on the event data that Playwright returns to the recorder.

When the client receives the event data, it will now use the supplied `actionTitles` field to display all the actions that the step contained.

1. Record a series of steps, preferably three or more.
2. Insert an assertion into the middle step that should cause the step to fail.
3. Test your script.
4. Observe that the first step shows succeeded, as per the screenshot pictured above. Observe the second step shows the usual failure information, along with the code. Observe that the remaining steps are shown as skipped, along with all of their action titles.

### DIFF
--- a/src/components/TestResult/ResultBody.tsx
+++ b/src/components/TestResult/ResultBody.tsx
@@ -22,29 +22,31 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 */
 
-import { EuiFlexItem, EuiText } from "@elastic/eui";
+import { EuiFlexGroup, EuiFlexItem, EuiText } from "@elastic/eui";
 import React from "react";
 import type { StepStatus } from "../../common/types";
-import { ResultContentWithoutAccordion, symbols } from "./styles";
+import { ResultContentWrapper, symbols } from "./styles";
 
 interface IResultBody {
-  durationElement: JSX.Element;
-  name: string;
+  actionTitles: string[];
   resultCategory: StepStatus;
 }
 
-export function ResultBody({
-  durationElement,
-  name,
-  resultCategory,
-}: IResultBody) {
+export function ResultBody({ actionTitles, resultCategory }: IResultBody) {
   return (
-    <ResultContentWithoutAccordion alignItems="center" gutterSize="xs">
-      <EuiFlexItem grow={false}>{symbols[resultCategory]}</EuiFlexItem>
-      <EuiFlexItem>
-        <EuiText size="s">{name}</EuiText>
-      </EuiFlexItem>
-      <EuiFlexItem grow={false}>{durationElement}</EuiFlexItem>
-    </ResultContentWithoutAccordion>
+    <EuiFlexGroup direction="column" gutterSize="none">
+      {actionTitles.map((name, index) => (
+        <ResultContentWrapper
+          alignItems="center"
+          key={name + index}
+          gutterSize="xs"
+        >
+          <EuiFlexItem grow={false}>{symbols[resultCategory]}</EuiFlexItem>
+          <EuiFlexItem>
+            <EuiText size="s">{name}</EuiText>
+          </EuiFlexItem>
+        </ResultContentWrapper>
+      ))}
+    </EuiFlexGroup>
   );
 }
--- a/e2e/tests/testResult.test.ts
+++ b/e2e/tests/testResult.test.ts
@@ -42,5 +42,6 @@ describe("Steps", () => {
     await electronService.clickRunTest();
 
     expect(await electronWindow.$("text=1 success"));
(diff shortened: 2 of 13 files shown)


src/components/TestResult/ResultBody.tsx
```diff
@@ -22,29 +22,31 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
...
+          key={name + index}
+          gutterSize="xs"
+        >
+          <EuiFlexItem grow={false}>{symbols[resultCategory]}</EuiFlexItem>
+          <EuiFlexItem>
+            <EuiText size="s">{name}</EuiText>
+          </EuiFlexItem>
+        </ResultContentWrapper>
+      ))}
+    </EuiFlexGroup>
```
lucasfcosta, line 50: Isn't this the same as the first `EuiFlexGroup` in `src/components/TestResult/ResultErrorBody.tsx`?

Shouldn't we reuse this down there?

lucasfcosta (approved):
Just tested this E2E and I've found a small problem: **the state of the dots on the left indicating the step success/failure seem to persist from one recording to another**.

How to reproduce:
1. Recording a journey with multiple steps and actions.
2. Adding an action in the middle of a step (the second, for example) which makes it fail. Confirm the dots on the left show as red.
3. Stop recording and start recording a new journey.
4. You'll see that the actions in the step that has previously failed still show up as red despite having never run.

See the GIF:

bad_state

lucasfcosta (changes requested):
(was meant to request changes in the previous review)

src/components/TestResult/ResultBody.tsx
justinkambic: Yes, it's the same exact props. In my opinion, it's not worth reusing because there's nothing novel here in the way the props are defined and there's no custom styling or other modification of the underlying functionality.

I think if the flex group's props were computed or if we had added some custom style it could be worthwhile to extract it and reuse it.

I don't think there's any major compute savings either, is there an aspect to this I'm overlooking?

justinkambic:
> Just tested this E2E and I've found a small problem: the state of the dots on the left indicating the step success/failure seem to persist from one recording to another.

Great find, thank you for pointing this out! Please have a look at https://github.com/elastic/synthetics-recorder/pull/159/commits/6d1587b477d7cbcd77d709a7d5cf10636bd759b5 when you have time, it should resolve that issue.

src/components/TestResult/ResultBody.tsx
lucasfcosta: No major compute savings indeed. This came to mind just to keep code "DRY" given it's the same markup and arguments. Given it's an extra indirection, however, I'm not strongly opinionated about making it a separate component unless we start using it more often.

lucasfcosta (approved):
LGTM, also tested E2E. Thanks for these changes 🙌

src/components/TestResult/ResultBody.tsx
justinkambic: I'm usually alright with having a single repetition elsewhere in the codebase. If we were to start doing this more than that, then I think it's worth maintaining a separate file for it. Just my opinion though.

PR #160, opened 2022-02-16
feat: Add tooltip to assertion element heading

### LINKED ISSUE #65: [Script Recorder] Assertion
As a user, I need to

- [x] add an assertion related to a specific action at any point of my journey
- [x] edit an assertion being able to update the assertion type, selector, and expected value
- [x] delete an assertion

to easily validate an action content of my flow

<details>
<summary>Deliverables</summary>

43Xhfql1cE

- [Figma](https://www.figma.com/file/6xfG7uIpgc84d62JnXhOMV/Script-recorder?node-id=774%3A128000)
- [Prototype](https://www.figma.com/proto/6xfG7uIpgc84d62JnXhOMV/Script-recorder?node-id=756%3A126651&scaling=scale-down&page-id=683%3A126181&starting-point-node-id=683%3A126878&show-proto-sidebar=1)
</details>

<details>
<summary>User feedback first iteration (archive)</summary>

## Feedback on the initial flow we showed to the users
We received clear feedback about the assertion after users added them.
Some points/considerations were raised during user testings:

- In the current build we call them Assertion 1,2,3... and we show the selector in the tooltip. (see screenshot). Calling them this way doesn't give any value. Let's remove it and show the selector field instead. 

Screenshot 2021-11-02 at 18 02 21

- Having the selector as a tooltip is too hidden, some users would like to see this upfront. We also have to truncate this if this is too long so my proposal is to replace Assertion 1,2,3 with the selector name/type.

Screenshot 2021-11-23 at 11 20 09 

- Once they added assertion users expect to be able to edit the selector field straight away without opening the side panel again (same functionality of editing the step or action) so I think we should reuse the same pattern we will need to use for editing the step name.
</details>

justinkambic:
## Summary

Resolves #65.

Adds a tooltip to the assertion element's heading per the designs.

image

## Implementation details

Note: it does not include a link as the design indicates. I'm not sure what location that link should go to, and the tooltip's default behavior is to collapse when the mouse is no longer hover on the child element.

We can modify this implementation by fleshing out this requirement more.

Record steps, insert an assertion, edit it and see the tooltip appear over the icon.

### DIFF
--- a/e2e/services/electron.ts
+++ b/e2e/services/electron.ts
@@ -94,6 +94,18 @@ export class ElectronServiceFactory {
     await electronWindow.click("text=Stop");
   }
 
+  async clickActionElementSettingsButton(
+    elementSelector: string,
+    buttonSelector: string
+  ) {
+    const electronWindow = await this.getWindow();
+    await electronWindow.hover(elementSelector);
+    await electronWindow.click(
+      `[aria-label="Expand the settings menu for this action"]`
+    );
+    return electronWindow.click(buttonSelector);
+  }
+
   async waitForPageToBeIdle(timeout = 45000) {
     await this.#recordingBrowserPage.waitForLoadState("networkidle", {
       timeout,
--- a/e2e/tests/assertionInfo.test.ts
+++ b/e2e/tests/assertionInfo.test.ts
@@ -1,0 +1,60 @@
+/*
+MIT License
+
+Copyright (c) 2021-present, Elastic NV
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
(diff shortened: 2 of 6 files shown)


vigneshshanmugam (approved):

lucasfcosta (approved):
LGTM, also validated E2E 🙌

liciavale (changes requested):
I had a chat with Justin about this and I think we need to replace the tooltip.
It should not be a tooltip but a popover and we should add the link that links to the [playwright doc ](https://playwright.dev/docs/assertions/) in case users want to learn more. 
Sorry to push this back.

justinkambic:
@liciavale can you comment on this implementation from a design perspective and let me know if there's anything you'd like to see adjusted?

 
image


Thanks for catching the incorrect element.

justinkambic:
@liciavale see this updated screenshot:

image

liciavale (approved):

PR #165, opened 2022-02-24
chore: Download browsers for all platforms to release builds from single machine

### LINKED ISSUE #150: [Other] Fix browser inclusion in binary for all platforms
### Task summary

Today, when we run our release script, it works quite well. Electron builder outputs binaries for Windows, Linux, and macOS.

Unfortunately, we have some custom scripting in our build process responsible for packaging the correct browser version as part of the process. This only works for the current platform though, so if you build all three binaries on a Mac, only the Mac version of the browser gets baked into the DMG. The Windows and Linux binaries will fail to run chromium in a recording session, because the browser is missing.

### Linked resources

The script responsible for installing Playwright: https://github.com/elastic/synthetics-recorder/blob/main/scripts/install-pw.js.

### Additional information

It's possible to work around this problem by building binaries for each platform natively. The drawback is it fragments the process and requires access to multiple build machines/CI processes.

### LINKED ISSUE #172: [Bug] Playback does not work on Linux
### Bug summary

When running my Synthetics Recorder in Ubuntu(18.04 and 20.04), I can successfully record my scripts, but when I press the button `Test`, nothing happens apart from the button getting disabled and enabled again.

We could observe some error messages from the app which indicates the browser fails to launch:
image

Frankly, the playback works when running it in headless mode(I commented out `no-headless` option when launching synthetics-cli)

Here are other insights from @emilioalvap:
> I've debugged on general X stuff on Ubuntu and Debian (there are some recent complaints about ubuntu wayland), it seems we are invoking two instances with different environment configurations. The first image is the invocation when user clicks "Start" and the second one is when the user clicks "Test". The environment in the second case is stripped of many of the variables that X uses to map apps to displays. (Sorry for the small images, I'm working with Hyper-V on a stripped machined and it doesn't allow resizing)
image
image

### Recorder Version

0.0.1-alpha.2

### Steps to reproduce

1. Start the recorder from the command-line tool, whether from prebuilt binary or in dev mode on Linux, for example, Ubuntu.
2. Record a simple scripts
3. Press `Test` button to start playback
4. See nothing happens on the Recorder, and see the error messages printed on your terminal

### Expected behavior

It should launch a browser and expect to see the playback running.

### Additional information

_No response_

kyungeunni:
## Summary
fixes #150
fixes #172 (fixed by 891b9f6)

This PR adds an additional pre-build step. When running `npm run release` script, the pre-build script will download browsers for Mac Windows, and Linux (if the browsers are already downloaded, it will skip downloading steps. The path contains the revision number so it also prevents packaging the stale browser.). Then electron builder will include the right browser for each platform as specified in `electron-builder.yml`.

## Implementation details
I couldn't utilize the playwright's registry library, as they are hardly coupled with the host machine's OS. Instead, I've used fetchBrowser library that is used by the registry's install function. 

1. run`npm run pack -- -mwl`. It will digest builds in your local `dist` directory without publishing it.
2. try run each of the binaries in each platform.
  - [x] Mac
  - [x] Windows
  - [x] Linux (`.deb` for the time being)

## Next Steps
- Release the binaries as described in [RELEASE.md](https://github.com/elastic/synthetics-recorder/blob/main/RELEASE.md)
- Stretch it to include Apple Silicon build
- Run the release script in CI on tags push.


justinkambic:
Thanks for taking this on, it represents a huge accomplishment.

In my process of smoke testing binaries I built with this branch, I encountered a problem running tests on Linux and Windows.

image

This problem doesn't happen on the Mac build, so I think we are probably seeing the same issue here as we had with the browsers. Sharp isn't getting copied over for non-native builds, and thus it seems to be causing the test feature to fail.

electron-builder.env
```diff
@@ -0,0 +1 @@
+BROWSER_REVISION=907428
```
vigneshshanmugam, line 1: We can automate this with ENV variable. We need to get it from https://github.com/elastic/synthetics-recorder/blob/13a146375d4582a04361d057969c3ab17c6e5e42/scripts/install-pw.js#L29

scripts/download-browsers.js
```diff
@@ -0,0 +1,97 @@
...
+
+const util = require("util");
+const path = require("path");
+const {
+  downloadBrowserWithProgressBar,
+} = require("playwright/lib/utils/browserFetcher");
+const { getChromeVersion } = require("./install-pw");
+
+const SUPPORTED_PLATFORMS = ["linux", "mac", "win"];
+const EXECUTABLE_PATHS = {
```
vigneshshanmugam, line 35: Need `ffmpeg` as well if i remember correctly for all platforms. Please double check if thats not the case.

scripts/download-browsers.js
```diff
@@ -0,0 +1,97 @@
...
+    );
+    try {
+      await download(platform, revision, directory);
+    } catch (e) {
+      throw Error("Failed to download browser for platform " + platform);
+    }
+  }
+}
+// eslint-disable-next-line no-console
+downloadAll().then(() => console.log("Download succeeded"));
```
vigneshshanmugam, line 97: Move them inside this check if we want to execute only when running as main module and not while importing - https://github.com/elastic/synthetics-recorder/blob/13a146375d4582a04361d057969c3ab17c6e5e42/scripts/install-pw.js#L50

package.json
```diff
@@ -12,9 +12,11 @@
...
     "electron:copy": "cp -r ./electron/. ./build",
     "dev": "concurrently -k \"BROWSER=none npm run react:start\" \"wait-on tcp:3000 && nodemon --exec electron electron/electron.js\"",
-    "build": "npm run clean && npm run react:build && npm run electron:copy",
-    "pack": "npm run build && electron-builder",
-    "release": "npm run build && electron-builder -mwl --publish=always",
+    "build": "run-s clean react:build electron:copy",
+    "release": "npm run pack -- -mwl --publish=always",
+    "pack": "run-s prepack electron-builder",
+    "prepack": "run-p build prepack:browsers",
+    "prepack:browsers": "node scripts/download-browsers",
```
vigneshshanmugam, line 19: As @justinkambic experienced, this would still break for native deps, We can fix it by running the preparation and package script serially for all platforms and uploading the pack one by one.

scripts/download-browsers.js
```diff
@@ -0,0 +1,97 @@
...
+};
+
+const CHROMIUM_DOWNLOAD_URLS = {
+  linux: "%s/builds/chromium/%s/chromium-linux.zip",
+  mac: "%s/builds/chromium/%s/chromium-mac.zip",
+  win: "%s/builds/chromium/%s/chromium-win64.zip",
+  // 'mac-arm64': '%s/builds/chromium/%s/chromium-mac-arm64.zip',
+};
+
+async function download(platform, revision, directory) {
```
vigneshshanmugam, line 48: Did you try to check the Registry in Playwright if there was any other way to do this?

vigneshshanmugam (commented):
Great work @kyungeunni 🎉

scripts/download-browsers.js
kyungeunni: Yes I went through all the Registry in Playwright and it seemed to be no work way around in there, especially I couldn't mock the platform as they would check the host machine's platform in various steps. So I used one of the functions that Registry uses to download the browser which is not tied to the host machine's platform(you pass it via parameter)

scripts/download-browsers.js
kyungeunni: Thanks for reminding me this!! I've updated the script to also download `ffmpeg` too.

kyungeunni:
This PR is ready for re-review. What I've fixed since the last review:
- Download `ffmpeg` as well as `chromium` before packing. Previously I've downloaded all the browsers for all the platform prior to packing, but I've updated it to download browsers per the requested platform. (e.g. when building binary for windows, download only for windows).
- Fixed native module download issue, namely `sharp` which is used by `@elastic/synthetics`. electron-builder supposed to handle this but it is incompatible with how sharp's install script works. more details are described [here](https://github.com/lovell/sharp/issues/2764). 

The latest build works well on Mac and Windows. When running it on Linux(Ubuntu 18.04) it had an issue launching the playback(test) but that is out of the scope as it doesn't seem to be related to the build. (I will create another ticket regarding that shortly)

cc @justinkambic @vigneshshanmugam

justinkambic:
Fantastic update @kyungeunni and thank you for the advice @emilioalvap ❤️ 

I will review as soon as possible.

kyungeunni:
Hey @lucasfcosta 👋 @justinkambic mentioned that it would be great if you could test running the build script from your machine and see if it works well just to make sure it is not customed to work only on my machine. When you have a minute, could you try running `npm run pack -- -mwl` and see all the builds are created under `/dist` without an error please? 🙏 I'm not sure

justinkambic:
@kyungeunni I'm still seeing an issue on Linux unfortunately 😞 

image

justinkambic:
@kyungeunni I'm still seeing an issue on Linux unfortunately 😞 

image

vigneshshanmugam:
Context - `.synthetics` folder gets created by Synthetics library to put the screenshots and other cached artifacts to be constructed later when the journeys are completed. We need to add permission to be able to create that folder across all OS.

kyungeunni:
Thanks, @vigneshshanmugam !! I'm wondering if it was a permission issue, shouldn't the error code be `EACCES: permission denied`?

scripts/beforePack.js
```diff
@@ -0,0 +1,57 @@
...
+
+const { Arch } = require("electron-builder");
+
+function fixSharp(arch, platform) {
+  return new Promise((resolve, reject) => {
+    const npmInstall = spawn("npm", ["run", "fix-sharp"], {
+      stdio: "inherit",
+      shell: true,
+      env: {
+        ...process.env,
```
lucasfcosta, line 41: Here too, I'm not sure if this is a big deal, but I'd imagine that if anyone pushes a malicious update to `libvips` or any other library we use in the `fix-sharp` command, they could have an environment variable containing our secrets used for signing the app or other credentials.

Maybe I'm being paranoid or there are mechanisms to prevent this already, but I'd appreciate more thoughts from others here.

scripts/download-browsers.js
```diff
@@ -0,0 +1,135 @@
...
+    case "darwin":
+      return "mac";
+    default:
+      return platform;
+  }
+}
+
+exports.downloadForPlatform = async function downloadForPlatform(platform) {
+  platform = translatePlatform(platform);
+  for (const browser of ["chromium", "ffmpeg"]) {
```
lucasfcosta, line 112: Nit: should we change `browser` to `software` here and in the references within the function like `getBrowserVersion` given we're not only dealing with browsers here?

lucasfcosta (commented):
Thanks for this @kyungeunni!

My only comments were around env vars and small nits.

Btw, I did build the runner on my Mac and ran the built recorder. I was able to record a multi-step journey and replay it just fine :)

scripts/beforePack.js
emilioalvap: @lucasfcosta Good point. I'm not sure we would like to follow an allow list approach here, since we are taking into consideration multiple types (X, chromium, playwright, distro, etc...) and user-defined variables. 
IMO, it should suffice to filter those variables that are only meant to be used in the recorder itself: user auth, tokens and such.

vigneshshanmugam:
> Thanks, @vigneshshanmugam !! I'm wondering if it was a permission issue, shouldn't the error code be EACCES: permission denied? 

It seems like we are trying to access it for the screenshots, In that case, can you try with `screenshots: off` when running Synthetics test command. That should make the issue go away.

lucasfcosta:
> > Thanks, @vigneshshanmugam !! I'm wondering if it was a permission issue, shouldn't the error code be EACCES: permission denied?
> 
> It seems like we are trying to access it for the screenshots, In that case, can you try with `screenshots: off` when running Synthetics test command. That should make the issue go away.

@vigneshshanmugam I think even if we turn it off, the runner would still try to create a folder as specified [here](https://github.com/elastic/synthetics/blob/201f7da7db8d8ce9c13855023c8566b73ebaf91a/src/core/runner.ts#L442-L458) though, wouldn't it? Or am I missing something?

scripts/beforePack.js
kyungeunni: Yeah, that's a valid point 👍 At the same time, I also thought would we have the same security concerns if we were to run the same script in the main process. Let me know if I'm being so naive!! Anyways It is always to be precautious about it, and I really like what @emilioalvap suggested. We should consider auditing some of the sensitive env vars and filter them out before passing to the child process.

scripts/download-browsers.js
kyungeunni: That's a good point, will amend 👍

scripts/beforePack.js
```diff
@@ -0,0 +1,57 @@
...
+    npmInstall.on("close", code => {
+      if (code === 0) {
+        resolve();
+      } else {
+        reject(new Error("process finished with error code " + code));
+      }
+    });
+    npmInstall.on("error", error => {
+      reject(error);
+    });
```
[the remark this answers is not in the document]
kyungeunni: Thanks updated (The less code the better 🧹)

vigneshshanmugam:
> @vigneshshanmugam I think even if we turn it off, the runner would still try to create a folder as specified [here](https://github.com/elastic/synthetics/blob/201f7da7db8d8ce9c13855023c8566b73ebaf91a/src/core/runner.ts#L442-L458) though, wouldn't it? Or am I missing something?

++ As you mentioned there is no way to turn it off, turning off screenshots wont just look for it.

electron/execution.js
```diff
@@ -245,6 +245,7 @@ async function onTest(data, browserWindow) {
      */
     synthCliProcess = fork(`${SYNTHETICS_CLI}`, args, {
       env: {
+        ...process.env,
```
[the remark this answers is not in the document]
justinkambic: This is something I'm uncertain about as I'm not accustomed to the security implications of changes to CLI interaction. From speaking offline it seems like this is not a large risk, but I'd like to hear others' opinions.

scripts/beforePack.js
```diff
@@ -0,0 +1,55 @@
...
+const { spawn } = require("child_process");
+const { downloadForPlatform } = require("./download-browsers");
+
+exports.default = function beforePack(ctx) {
+  const arch = Arch[ctx.arch];
+  const platform = ctx.electronPlatformName;
+  return Promise.all([downloadForPlatform(platform), fixSharp(arch, platform)]);
+};
+
+const { Arch } = require("electron-builder");
```
justinkambic, line 33: Is there a functional reason that `Arch` needs to be required after `beforePack` is initialized?

justinkambic (commented):
I'm going to comment here and leave a passive LGTM - we should make sure that we don't leave these threads questioning security topics open before merging. I'll reach out to the other reviewers to see if they have any further thoughts on revisions we can make here, or if they're alright to merge as-is.

electron/execution.js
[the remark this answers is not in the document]
kyungeunni: As @emilioalvap mentioned in another comment, I think there will be various types of env values as well as platform specific ones, so it can be a bit tricky to track all of them. We should measure the effort vs worth by doing so (if we decide to do it). 

Another note, when env option is not specified when spawning, Node assigns parent's `process.env` as a [default](https://nodejs.org/api/child_process.html#child_processspawncommand-args-options)

scripts/beforePack.js
lucasfcosta: Agreed, a filter _out_ should be enough rather than a filter _in_ just to be sure we're not passing our own signing keys to a malicious package downwards.

lucasfcosta (approved):
I'll follow @justinkambic's on this and leave this as a passive LGTM considering that filtering out the environment variables with our keys should be enough to prevent it from being passed downwards to other packages, although I do think we should just consult with security to be sure before merging.

As a note, I've also tested E2E the build for my MacOS and it worked fine for recording, replaying, and generating code for a journey.

andrewvc:
Works fine on windows!

scripts/afterPack.js
```diff
@@ -0,0 +1,52 @@
...
+THE SOFTWARE.
+*/
+const { spawn } = require("child_process");
+const path = require("path");
+
+exports.default = function afterPack(ctx) {
+  const platform = ctx.electronPlatformName;
+  if (platform === "linux") {
+    const resourcesPath = path.join(ctx.appOutDir, "resources");
+    changePermission(resourcesPath);
```
vigneshshanmugam, line 31: Did you miss the return statement here?

vigneshshanmugam (approved):
LGTM, Great work @kyungeunni

scripts/beforePack.js
andrewvc: Just so I can better understand here, isn't there a whole stack of dependencies we already use, given that we use an OSS toolchain? I guess my question is why are we particularly concerned here? I spoke briefly with @emilioalvap , and he pointed out that NPM is perhaps weaker than a linux distro in terms of supply chain (though probably not for many attacks). 


CC @levinebw, your thoughts here?

scripts/beforePack.js
```diff
@@ -0,0 +1,54 @@
...
+THE SOFTWARE.
+*/
+const { spawn } = require("child_process");
+const { Arch } = require("electron-builder");
+const { downloadForPlatform } = require("./download-browsers");
+
+exports.default = function beforePack(ctx) {
+  const arch = Arch[ctx.arch];
+  const platform = ctx.electronPlatformName;
+  return Promise.all([downloadForPlatform(platform), fixSharp(arch, platform)]);
```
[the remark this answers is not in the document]
vigneshshanmugam: If the error handling is taken care of the electron-builder, then all good 👍🏽 I just wanted us to avoid UnhandledPromiseRejection errors.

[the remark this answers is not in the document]
kyungeunni: image
yeah it's handled by electron-builder and looks like this 👍 thanks for double checking it with me!!

lucasfcosta:
Just adding an approval here considering the conversation we've just had with someone representing the security team (meaning we won't do any signing at this stage and thus not pass credentials down through `env`).

PR #176, opened 2022-03-23
fix: Update EUI version and switch to Amsterdam

### LINKED ISSUE #167: [Bug] Use full Amsterdam theme
### Bug summary

Recently, the EUI squad helpfully pointed out that the recorder's using the older EUI theme rather than the latest Amsterdam one. We should work with them to get the recorder running the latest versions of the theme.

The [have documented](https://elastic.github.io/eui/#/guidelines/getting-started) a getting started experience, which we can use as a baseline for making sure our usage of the framework is in line with their expectations.

### Recorder Version

latest

### Steps to reproduce

Run the recorder, see outdated UI.

### Expected behavior

Run the recorder, see Amsterdam UI.

### Additional information

_No response_

justinkambic:
## Summary

Resolves #167.

As noted in the linked issue, we aren't using the latest and greatest EUI theme. Fixing that here.

## Implementation details

There may be minor things that break and are tough to notice. It might be worthwhile to build a binary from `main` and compare it against this branch in dev mode.

There should be no functional change with this patch, it should update EUI and utilize the Amsterdam styles.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -25,6 +25,7 @@ THE SOFTWARE.
 import React, { useContext } from "react";
 import { useEffect, useState } from "react";
 import { EuiCode, EuiEmptyPrompt, EuiProvider } from "@elastic/eui";
+import createCache from "@emotion/cache";
 import "./App.css";
 import "@elastic/eui/dist/eui_theme_light.css";
 import { Title } from "./components/Header/Title";
@@ -39,7 +40,6 @@ import { useSyntheticsTest } from "./hooks/useSyntheticsTest";
 import { generateIR, generateMergedIR } from "./helpers/generator";
 import { StepSeparator } from "./components/StepSeparator";
 
-import "@elastic/eui/dist/eui_theme_light.json";
 import "./App.css";
 import { useStepsContext } from "./hooks/useStepsContext";
 import { TestResult } from "./components/TestResult";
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -24,14 +24,10 @@ THE SOFTWARE.
...
 import React, { useContext } from "react";
 import { useEffect, useState } from "react";
-import {
-  EuiCode,
-  EuiEmptyPrompt,
-  EuiThemeProvider,
-  EuiThemeAmsterdam,
-} from "@elastic/eui";
+import { EuiCode, EuiEmptyPrompt, EuiProvider } from "@elastic/eui";
+import createCache from "@emotion/cache";
```
vigneshshanmugam, line 28: I don't see the @emotion/cache on the deps list.

src/App.tsx
```diff
@@ -54,6 +49,13 @@ import { ExportScriptFlyout } from "./components/ExportScriptFlyout";
 import { useRecordingContext } from "./hooks/useRecordingContext";
 import { StartOverWarningModal } from "./components/StartOverWarningModal";
 
+const cache = createCache({
```
vigneshshanmugam, line 52: can you add some comments on how its used and whats its for?

src/App.tsx
justinkambic: We already had it as a dep of `@emotion/react`, which I guess is why it worked. I've added it as a direct dep. d8e107d7f2e77f14e8ae730e55a06162befeb814

src/App.tsx
justinkambic: Basically there are issues with EUI that necessitate this. I've added a comment and a link to the relevant documentation. 0d71c584cfc307996d086f1e261fdaa818368723

vigneshshanmugam: Thanks they helps

vigneshshanmugam (approved):
LGTM

PR #177, opened 2022-03-24
chore: update deps and add `lint-staged`

justinkambic:
## Summary

The pre-commit hook aspect of the repo leaves a lot to be desired. I've added `lint-staged` which greatly improved the `lint:fix` functionality in terms of making commits to the repo.

Additionally, I've updated several dependencies that could use a bump, and installed missing peer dependencies to remove warning messages.

## Implementation details

There should be no functional change to the recorder.

`lint-staged` should kick in when attempting to commit changes.

Ensure the recorder builds/runs ok. Passing CI should also be a positive indicator that this patch is alright.

Make a lint error such as removing one of the copyright headers from a source file, stage it and attempt to commit. The file should automatically be updated to include the header as part of the commit. Don't push test commits.

### DIFF
--- a/demo-app/package.json
+++ b/demo-app/package.json
@@ -10,7 +10,7 @@
   "dependencies": {
     "final-form": "^4.20.1",
     "js-cookie": "^2.2.1",
-    "next": "10.0.3",
+    "next": "12.1.0",
     "react": "17.0.1",
     "react-dom": "17.0.1",
     "react-final-form": "^6.5.2",
--- a/package.json
+++ b/package.json
@@ -82,6 +82,17 @@
       "category": "Utility"
     }
   },
+  "lint-staged": {
+    "*.{js,jsx,ts,tsx}": [
+      "npm run lint:fix",
+      "prettier --write"
+    ]
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
   "devDependencies": {
     "@testing-library/react": "^12.1.2",
     "@testing-library/react-hooks": "^7.0.2",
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -82,22 +82,34 @@
...
+    "*.{js,jsx,ts,tsx}": [
+      "npm run lint:fix",
+      "prettier --write"
+    ]
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
```
kyungeunni, line 95: I might be wrong, but is this redundant since we have `.husky/pre-commit` file?

justinkambic: Right you are. https://github.com/elastic/synthetics-recorder/pull/177/commits/b65e0afebb26740796178727f9b7349e3705c403

demo-app/package.json
```diff
@@ -10,7 +10,7 @@
   "dependencies": {
     "final-form": "^4.20.1",
     "js-cookie": "^2.2.1",
-    "next": "10.0.3",
+    "next": "12.1.0",
```
vigneshshanmugam, line 13: why is next.js in deps?

vigneshshanmugam (approved):
LGTM except for the next.js stuff.

kyungeunni (approved):
LGTM :shipit:

demo-app/package.json
vigneshshanmugam: Oops, forgot that it was part of Demo app json. Ignore me. Also did we check if that demo app actually works?

justinkambic: I just ran the demo app. Seems to work fine.

Issue #161, opened 2022-02-23
[Bug] URL should increase when re-sizing the window


liciavale:
### Bug summary

When users increase the size of the recorder, The URL should be responsive too

<img width="1562" alt="Screenshot 2022-02-22 at 14 01 29" src="https://user-images.githubusercontent.com/8301088/155313106-c6cc4a24-a196-440f-a136-3d2860f83666.png">

### Steps to reproduce

1. Open the script recorder
2. Increase the size of the window

### Expected behavior

The URL should be responsive

emilioalvap:
Tested on v0.0.1-alpha.3, LGTM

PR #178, opened 2022-03-28
fix: URL input expands for longer value

justinkambic:
## Summary

Resolves https://github.com/elastic/synthetics-recorder/issues/161.

## Implementation details

Previously the URL input would not expand to fill available space. This PR makes the input maintain its original size, unless the user begins to enter an excessively long URL. In that case, the input field will expand to the bounds of the parent element.

Additionally, the `grow` prop for the recording controls is now dependent on the URL. It will `grow` for small URLs, but to maximize space usage it will stop `grow`ing so the URL field can fill the entirety of the available whitespace in the header controls area.

Check that a smaller URL value results in the field staying small.

Check that a larger URL results in the field filling the available space in the header control section.

### DIFF
--- a/src/components/Header/UrlField.tsx
+++ b/src/components/Header/UrlField.tsx
@@ -28,6 +28,7 @@ import type { Setter } from "../../common/types";
 import { RecordingStatus } from "../../common/types";
 
 export interface IUrlField {
+  fullWidth?: boolean;
   recordingStatus: RecordingStatus;
   setUrl: Setter<string>;
   toggleRecording: () => void;
@@ -38,6 +39,7 @@ export const URL_FIELD_LABEL =
   "Set the starting point for your synthetic journey";
 
 export function UrlField({
+  fullWidth,
   recordingStatus,
   setUrl,
   toggleRecording,
@@ -46,6 +48,7 @@ export function UrlField({
   return (
     <EuiFieldText
       aria-label={URL_FIELD_LABEL}
+      fullWidth={fullWidth}
       onChange={e => {
         setUrl(e.target.value);
       }}
--- a/src/components/Header/UrlField.tsx
+++ b/src/components/Header/UrlField.tsx
@@ -28,7 +28,6 @@ import type { Setter } from "../../common/types";
 import { RecordingStatus } from "../../common/types";
 
 export interface IUrlField {
-  fullWidth?: boolean;
   recordingStatus: RecordingStatus;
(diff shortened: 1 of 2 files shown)


lucasfcosta (changes requested):
This functionality looked a bit weird to me as the field instantly grows once a certain character limit is exceeded. That looks weird to me because if the user has a larger recorder window, there will just be a lot of empty space between the field and the button. Then, when an arbitrary number of characters is reached, the field will expand and may either:

A) Become much bigger than it's necessary
B) Not become large enough if the user pastes an excessively long URL

IMO always turning the `flex-grow` "on" for this field would make it behave more consistently so that users don't see this "flickering" and so that the input always stays next to start. Furthermore this PR becomes essentially a ~2 lines PR, simplifying things.

I've attached a GIF to illustrate what I mean:

url_expander

justinkambic:
Hey @lucasfcosta thanks for your feedback. I will give your proposal a try as I agree it's more consistent. Just to give you a window into my thinking when I wrote this initial approach - I assumed that users will typically be pasting in long URLs, but typing it is of course a valid use case as well and this solution is poorly adapted to covering it.

justinkambic:
@lucasfcosta please give this another look and LMK what you think

lucasfcosta:
Hey Justin, thanks for the update. That makes a lot of sense, indeed I don't imagine anyone will be typing long URLs, although pasting on a big window would cause the weird effect (but less noticeably as you'll be pasting only once.

It LGTM now with both short and long URLs :)

Screenshot 2022-03-31 at 09 35 07

Screenshot 2022-03-31 at 09 35 21

lucasfcosta (approved):

src/components/Header/UrlField.tsx
```diff
@@ -46,6 +48,7 @@ export function UrlField({
   return (
     <EuiFieldText
       aria-label={URL_FIELD_LABEL}
+      fullWidth={fullWidth}
```
lucasfcosta, line 51: Nit: the header is the only place we use`UrlField` so instead of having `fullWidth` as a prop we can simply remove it as a prop from `UrlField` itself and make the `EuiFieldText` always `fullWidth`.

justinkambic: Fixed: https://github.com/elastic/synthetics-recorder/pull/178/commits/8a74b840967bd9f0992e96b48d0125335b3a98ce

PR #179, opened 2022-03-28
fix: Make assertions expanded on init

justinkambic:
## Summary

Resolves #162.

Assertions are expanded when created.

## Implementation details

I've updated the inline state management for assertion elements to start out as expanded.

## How to validate this change

1. Record some actions
1. Add an assertion after one of your actions
1. Observe that the assertion you added is expanded when created

PR #182, opened 2022-03-30
fix: Fill defect for `ControlButton`

justinkambic:
## Summary

As part of #179 I merged a defect that caused a fill to be applied to all instances of `ControlButton`. This patch fixes it to restore the intended functionality.

### Before
image

### After
image

## Implementation details

Changes the `fill` prop to be conditional again.

Make sure that not all of the buttons have `fill`s.


kyungeunni (approved):
Thanks Justin 🥇

PR #187, opened 2022-04-05
chore: use same prettier setting as Kibana's

### LINKED ISSUE #151: [Other] Use the same prettier config as Kibana
### Task summary

To avoid having a different coding style within the organization(read more about it [here](https://www.smashingmagazine.com/2012/10/why-coding-style-matters/)), I'd like to suggest using the same prettier config as Kibana's config.

It is going to be a simple task but considering there are many PRs opened at the moment, and running lint --fix would potentially cause some conflicts, I'd recommend we do it when the biggest PRs are merged.

### Linked resources

Kibana's prettierrc: https://github.com/elastic/kibana/blob/main/.prettierrc

### Additional information

_No response_

kyungeunni:
closes #151

Copied over `.prettierrc` from Kibana so that we align code style across the team within the organization.

No need to review every single file that has changed, but take a look at `.prettierrc` file.


e2e/services/browser.ts
```diff
@@ -22,12 +22,12 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
...
 
-import { chromium, Browser } from "playwright";
-import { env } from "../services";
+import { chromium, Browser } from 'playwright';
+import { env } from '../services';
 
 type ConnectRetryParams = { url?: string; timeout?: number; interval?: number };
 
-const sleep = ms => new Promise(resolve => setTimeout(resolve, ms));
+const sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms));
```
justinkambic, line 30: Personally, I kinda like omitting the parens if the type is implicit, but I'm fine with staying unified with Kibana's `.prettierrc`.

justinkambic (approved):
LGTM

e2e/services/browser.ts
kyungeunni: Hmm actually I prefer the same 🤔

demo-app/components/Ad.js
```diff
@@ -1,15 +1,10 @@
-import React from "react";
+import React from 'react';
```
lucasfcosta, line 1: IMO we should avoid linting the `demo-app` itself as that's a separate project we're just bundling in. It seems to me that's unnecessary to lint it given it's only used for testing and it's maintained separately. That would also make linting quicker.

lucasfcosta (approved):
LGTM but tbh I'd prefer to avoid linting the demo-app if possible given we just bundled it in from its own repo. No big deal though.

demo-app/components/Ad.js
kyungeunni: Also updated .eslintrc too

lucasfcosta (approved):
LGTM now that we're ignoring the `demo-app` 😄

kyungeunni:
closing in favor of #197.
Figured it is better to start from a clean state rather than resolving conflicts.

PR #189, opened 2022-04-05
chore: Remove unused exports

### LINKED ISSUE #188: [Other] Remove unused exports
### Task summary

We have a number of unused exports in the recorder codebase. We should write a script we can include in CI to disallow the inclusion of unused exports. This should be easy enough to remedy; the recorder is not a library and should not be exporting any module that is not used somewhere else in the application.

We should write a script and include it in `package.json`, and add it to our linting step, or create a new CI step that incorporates this check.

### Linked resources

I've done investigation into https://github.com/pzavolinsky/ts-unused-exports; it seems trivial to set up some basic tooling.

### Additional information

_No response_

justinkambic:
## Summary

Resolves #188.

We have a lot of unused exports in the recorder, our current linting setup does not catch these.

This enabled me to delete a bunch of dead code.

## Implementation details

This adds a new dev dependency, `ts-unused-exports`, and a new script that calls it. It can be run using `npm run unused-exports`. The script throws an error when unused exports are detected.

This will have a slightly annoying side effect of calling out exports that aren't referenced outside of the immediate file, but _do_ have references in the same file. In this case these symbols shouldn't _technically_ be exported, but are still used internally. This rule might be a little draconian, but I think it's worthwhile to prevent the existence of entire files and other objects that are completely dead.

I'm open to improvement suggestions, if people think this is too strong of an imposition.

You can create an unused export in any `ts`/`tsx` file, like `export type Unused = "nothing"` and run the script, it should throw with a reference to the filename and offending line.

I've also updated the CI to run this script, I will be debugging that before requesting review from others.


.ci/Jenkinsfile
```diff
@@ -61,6 +61,13 @@ pipeline {
             }
           }
         }
+        withGithubNotify(context: "Unused exports") {
+          withNodeJSEnv() {
+            dir("${BASE_DIR}"){
+              sh(label: 'npm unused-exports', script: 'npm run unused-exports')
+            }
+          }
+        }
```
kyungeunni, line 70: Can we run `npm run unused-exports` in pre-commit? What I expect here is that we will need to wait until the Jenkins job to fail and then compose another commit fixing it.

src/common/shared.ts
```diff
@@ -70,19 +70,6 @@ export const PLAYWRIGHT_ASSERTION_DOCS_LINK =
 
 export const SMALL_SCREEN_BREAKPOINT = 850;
 
-export function performSelectorLookup(
```
kyungeunni, line 73: Nice to see it detecting unused functions! 🔫

package.json
```diff
@@ -18,6 +18,7 @@
     "clean": "rm -rf ./build && rm -rf ./dist",
     "lint": "eslint .",
     "format": "prettier --write .",
+    "unused-exports": "node scripts/find-unused-exports.js",
```
kyungeunni, line 21: Not related to this PR but I'm a bit concerned to see the number of scripts in `package.json`. It is a lot (24 for now). It might be a good idea to plan a spring cleaning.

.ci/Jenkinsfile
justinkambic: Yes that's an improvement I wanted to add, I just didn't have time to get it tested and working when I pushed this. I'll address this later today if I can.

justinkambic: @kyungeunni can you please test that this is working now? I attempted to add it in https://github.com/elastic/synthetics-recorder/pull/189/commits/ad230e07b3fbe4d5f842dbd735b1b90387f995ac, and it looks like it works for me!

package.json
justinkambic: I'm not opposed to it. Feel free to create a ticket and re-organize/eliminate what you think we can.

kyungeunni (approved):
LGTM!

.ci/Jenkinsfile
justinkambic: We've discussed this offline but I'll respond here as well for posterity. We have decided that yes, we do want to run this on CI as well. It is possible to skip pre-commit hooks locally. If we don't perform a complete lint of the codebase as part of our integration process, it's possible that unused exports will be missed. And since the pre-commit hook uses `lint-staged`, we are never performing a full local check of the codebase, so when unused exports are merged they will never be caught, and we'll be back in the same situation in which we find ourselves.

By keeping this in CI, even if a contributor uploads an unused export, it will still be caught by the build machine, and we can avoid merging it as part of code review.

Issue #196, opened 2022-04-13
[Other] Add docs explaining how to avoid errors with custom step assignment


justinkambic:
### Task summary

As part of https://github.com/elastic/synthetics-recorder/issues/195, we should add documentation of the bug described by that issue, at least until we're able to resolve the bug.

PR #198, opened 2022-04-13
chore: Add update to `README.md` to explain undefined reference error

justinkambic:
## Summary

Resolves https://github.com/elastic/synthetics-recorder/issues/196.

Adds an update to the project's README that explains an error that can happen now that users are allowed to define and arrange their own step separators.

## Implementation details

Just a docs update.

Have a look at the example code and make sure it all makes sense. This code came directly from a real journey I ran, so it should be alright, but weirdness can happen.

Also find any typos, help eliminate passive voice, etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,52 @@ Run the recorder app in dev mode.
 npm run dev
 ```
 
+### Troubleshooting
+
+#### Page object is not defined
+
+It is possible to define a series of steps that splits your recorded actions in a way that causes the required
+page object to be out of scope. If you're seeing an error message like:
+
+```javascript
+page1 is not defined
+```
+
+You may be encountering this issue. You can verify it by clicking the `Export` button in the recorder and
+checking the code output. If your generated code is similar to this below, you may need to rearrange your steps.
+
+```javascript
+step("Click text=Babel Minify", async () => {
+  // `page2` is defined here
+  const [page2] = await Promise.all([
+    page.waitForEvent("popup"),
+    page.click("text=Babel Minify"),
+  ]);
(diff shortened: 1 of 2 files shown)


justinkambic:
> My only suggestion (non blocking) is that maybe the Possible install issues section should go as a #### heading, higher up under ### Installation and Usage

Sure I'm fine adding that change.

README.md
```diff
@@ -57,6 +57,53 @@ Run the recorder app in dev mode.
 npm run dev
 ```
 
+### Troubleshooting
+
+#### Page object is not defined
+
+It is possible to define a series of steps that splits your recorded actions in a way that causes the required
+page object to be out of scope. You may be encountering this issue if, when testing or running your journey, you
```
vigneshshanmugam, line 65:
```suggestion
page object to be out of scope in the generated code. You may be encountering this issue if, when testing or running your journey, you
```

README.md
```diff
@@ -57,6 +57,53 @@ Run the recorder app in dev mode.
...
+  ]);
+  await page2.click(':nth-match(a:has-text("babel-minify"), 3)');
+});
+step("Close page", async () => {
+  // referencing `page2`, which is now out of scope
+  await page2.close();
+});
+```
+
+By removing the second step divider in the Script Recorder UI, we will generate code that keeps the `page2` object
```
vigneshshanmugam, line 90:
```suggestion
By removing the step divider associated with `Close page` step in the Script Recorder UI, we will generate code that keeps the `page2` object
```

vigneshshanmugam (approved):
LGTM

kyungeunni (approved):
LGTM :shipit:

PR #204, opened 2022-04-13
fix: Add retry utility to action values test

### LINKED ISSUE #200: [Bug] Fix flaky e2e test
### Bug summary

We recently added [a test](https://github.com/elastic/synthetics-recorder/blob/main/e2e/tests/actionValues.test.ts) to ensure that edited values for actions and assertions are reflected in the outputted code. Unfortunately, on CI we have encountered some consistent flakiness/failure of this test, because the state update seems to take a bit longer on CI.

Example test output:

```
10:59:09  FAIL e2e/tests/actionValues.test.ts (6.427 s)
10:59:09    ● Assertion and Action values › includes updated action/assertion values in code output
10:59:09  
10:59:09      expect(received).toContain(expected) // indexOf
10:59:09  
10:59:09      Expected substring: "await page.goto('https://e2e-test-action-edit.org/');"
10:59:09      Received string:    "step('Go to http://localhost:11337/', async () => {·
10:59:09        await page.goto('http://localhost:11337/');·
10:59:09        expect(await page.innerText('test-assertion-selector')).toMatch('test-assertion-value');·
10:59:09      });"
10:59:09  
10:59:09        84 |       `expect(await page.${ACTION_OPTION}('${ASSERTION_SELECTOR}')).toMatch('${ASSERTION_VALUE}');`
10:59:09        85 |     );
10:59:09      > 86 |     expect(innerText).toContain(`await page.goto('${ACTION_URL}');`);
10:59:09           |                       ^
10:59:09        87 |   });
10:59:09        88 | });
10:59:09        89 |
10:59:09  
10:59:09        at Object.<anonymous> (e2e/tests/actionValues.test.ts:86:23)
10:59:09  
10:59:09  Test Suites: 1 failed, 1 of 6 total
10:59:09  Tests:       1 failed, 1 total
10:59:09  Snapshots:   0 total
10:59:09  Time:        6.492 s
```

I've verified the cause of this issue by adding an `await` that lasts about five seconds between the state update call and the attempt to fetch the code output.

image

### Recorder Version

`alpha-3`

### Steps to reproduce

[This line](https://github.com/elastic/synthetics-recorder/blob/main/e2e/tests/actionValues.test.ts#L78) is what's causing the problem. It occurs because Playwright clicks on the `Export` button in the Recorder UI faster than the state update from the previous actions can complete. Basically, we need to wait for this to finish.

### Expected behavior

The test should pass on CI as it does locally.

### Additional information

The original failure that led to the creation of this issue is ephemeral, but here's the link until the log is deleted: https://apm-ci.elastic.co/job/apm-agent-rum/job/synthetics-recorder-mbp/job/PR-198/2/execution/node/220/log/

justinkambic:
## Summary

Resolves #200.

Adds a retry mechanism to the action/assertion edit e2e test.

## Implementation details

The goal here is to ensure we allow adequate time for the state to update after saving changes. In the past playwright has managed to grab the outputted code string before the last update to the action is reflected.

CI should pass

### DIFF
--- a/e2e/tests/actionValues.test.ts
+++ b/e2e/tests/actionValues.test.ts
@@ -65,6 +65,45 @@ async function editAction(electronWindow: Page) {
   await electronWindow.click(`[data-test-subj="save-action-0-0"]`);
 }
 
+/**
+ * Utility that allows us to retry a step that may take some time after a previous action
+ * before it will be ready.
+ * @param assertionFunction the function that makes the assertions
+ * @param resetFunction an optional function to reset the UI state to a position ready for testing
+ * @param maxRetries the maximum number of times to retry before throwing an error
+ * @param sleepDuration the duration in ms to wait before trying the assertion function again
+ */
+async function retryAssertion(
+  assertionFunction: () => Promise<void>,
+  resetFunction?: () => Promise<void>,
+  maxRetries = 5,
+  sleepDuration = 1000
+) {
+  for (let i = 1; i <= maxRetries; i++) {
+    try {
+      await assertionFunction();
+    } catch (e) {
+      if (i === maxRetries) throw e;
+
+      console.log(
+        `Encountered assertion error, awaiting ${
+          sleepDuration / 1000
+        }s before retry. This is retry ${i + 1} of ${maxRetries}.`
+      );
+
+      // give time for state to change
+      await new Promise(r => setTimeout(r, sleepDuration));
+
+      // perform any necessary action to re-create the test state
+      if (resetFunction) await resetFunction();
+
+      // try again
+      continue;
+    }
+    break;
+  }
+}
+
 describe('Assertion and Action values', () => {
   it('includes updated action/assertion values in code output', async () => {
     const electronWindow = await electronService.getWindow();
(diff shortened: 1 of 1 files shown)


e2e/tests/actionValues.test.ts
```diff
@@ -72,8 +72,12 @@ describe('Assertion and Action values', () => {
     await editAssertion(electronWindow);
     await editAction(electronWindow);
 
+    // test fails without waiting for state to update
+    await new Promise(r => setTimeout(r, 1000));
```
kyungeunni, line 76:
```suggestion
  await electronWindow.waitForSelector(`id=action-element-0-0 >> div:has-text("${ACTION_URL}")`, { timeout: 2000 });

```

Maybe we could do something like this: It waits until the element is visible with the given selector. The selector I specified here contains the new url for the action. When the url is updated for the action, this line will resolve and we can go ahead and test the export feature (this is based on the understanding that we can say _the state is updated when the new value is visible in the Action header_)

We can move this line to `editAction` function as well so as it doesn't distract future readers.

** This will fail for now as we have a bug in the main branch that it shows the wrong title in the action title**

kyungeunni: https://github.com/elastic/synthetics-recorder/pull/212

I applied this change as part of bug fix. I will try to run several tests tomorrow so we can see if it's an effective fix or not.

justinkambic:
Closing this for now, in favor of https://github.com/elastic/synthetics-recorder/pull/212 which fixes the same problem.

PR #205, opened 2022-04-14
chore: convert electron js to ts

### LINKED ISSUE #141: [Enhancement] Convert Electron backend to TypeScript
### Feature summary

When I contribute code to the backend processes for the recorder, I want to use TypeScript.

As we are starting to introduce more contributors to this project, it will be less prone for us to introduce bugs and regressions if we are using TS on the backend.

### Linked resources

_No response_

### Additional information

_No response_

kyungeunni:
Resolves #141

// To be updated


electron/execution.ts
```diff
@@ -56,7 +59,7 @@ async function launchContext() {
   }
 
   context.on('page', page => {
-    page.on('dialog', () => {});
+    // page.on('dialog', () => {});
```
kyungeunni, line 62: I don't know what it does but TS complains as the callback is empty. Can we remove this listner on 'dialog' event?

electron/electron.ts
```diff
@@ -22,19 +22,19 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
...
+import isDev from 'electron-is-dev';
+import unhandled from 'electron-unhandled';
+import debug from 'electron-debug';
+import logger from 'electron-log';
+import setupListeners from './execution';
+import { buildMenu } from './menu';
 // For dev
 unhandled({ logger: err => logger.error(err) });
-debug({ enabled: true, showDevTools: false });
+debug({ isEnabled: true, showDevTools: false });
```
justinkambic, line 35: This is why we use TypeScript 😄

electron/execution.ts
justinkambic: I was able to record a journey and e2e tests are passing, so I think no problems here.

electron/execution.ts
```diff
@@ -103,7 +107,9 @@ async function recordJourneys(data, browserWindow) {
     const closeBrowser = async () => {
       browserContext = null;
       actionListener.removeListener('actions', actionsHandler);
-      await browser.close().catch({});
+      await browser.close().catch(err => {
```
justinkambic, line 110: Why don't we go ahead and use this syntax instead of combining `async...await` with callbacks.

```suggestion
      try {
        await browser.close();
      } catch (e) {
        logger.error("Browser close threw an error", e);
      }
```

electron/execution.ts
```diff
@@ -123,7 +129,7 @@ async function recordJourneys(data, browserWindow) {
  * @param {*} event the result data from Playwright
  * @returns the event data combined with action titles in a new object
  */
-function addActionsToStepResult(steps, event) {
+function addActionsToStepResult(steps: Steps, event: any) {
```
justinkambic, line 132:
```suggestion
function addActionsToStepResult(steps: Steps, event: TestEvent) {
```

This must be `TestEvent`, no?

electron/execution.ts
```diff
@@ -136,13 +142,15 @@ function addActionsToStepResult(steps, event) {
...
+        (action: ActionInContext, index: number) => action?.title ?? `Action ${index + 1}`
+      ),
     },
   };
 }
 
-async function onTest(data, browserWindow) {
-  const parseOrSkip = chunk => {
+async function onTest(data: RunJourneyOptions, browserWindow: BrowserWindow) {
+  const parseOrSkip = (chunk: string) => {
```
justinkambic, line 153:
```suggestion
  const parseOrSkip = (chunk: string): Array<Record<string, unknown>> => {
```

I think we can annotate the return type as this without any bad result.

electron/execution.ts
```diff
@@ -191,11 +199,11 @@ async function onTest(data, browserWindow) {
     }
   };
 
-  const sendTestEvent = event => {
+  const sendTestEvent = (event: TestEvent) => {
     browserWindow.webContents.send('test-event', event);
   };
 
-  const emitResult = chunk => {
+  const emitResult = (chunk: any) => {
```
justinkambic, line 206:
```suggestion
  const emitResult = (chunk: string) => {
```

I think `chunk` is a `string`, right?

electron/execution.ts
```diff
@@ -155,7 +163,7 @@ async function onTest(data, browserWindow) {
   };
 
   // returns TestEvent interface defined in common/types.ts
-  const constructEvent = parsed => {
+  const constructEvent: (parsed: Record<string, any>) => TestEvent | undefined = parsed => {
```
justinkambic, line 166:
```suggestion
  const constructEvent: (parsed: Record<string, unknown>) => TestEvent | undefined = parsed => {
```

I'd rather see us employ `unknown` here. We can define a [type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards) to check for the appropriate fields, or change the `switch` block to an `if...else` chain that can check for the appropriate fields that provides the correct type annotation.

electron/execution.ts
```diff
@@ -206,15 +214,16 @@ async function onTest(data, browserWindow) {
...
+  let synthCliProcess: ChildProcess | null = null; // child process, define here to kill when finished
   try {
     const isSuite = data.isSuite;
     const args = ['--no-headless', '--reporter=json', '--screenshots=off', '--no-throttling'];
     const filePath = join(JOURNEY_DIR, 'recorded.journey.js');
     if (!isSuite) {
       args.push('--inline');
     } else {
-      await mkdir(JOURNEY_DIR).catch(() => {});
+      // es-lint ignore
```
justinkambic, line 225: Explain why

electron/execution.ts
```diff
@@ -261,30 +271,31 @@ async function onTest(data, browserWindow) {
...
 
-async function onSetMode(mode) {
+async function onSetMode(mode: string) {
   if (!browserContext) return;
   const page = browserContext.pages()[0];
   if (!page) return;
   await page.mainFrame().evaluate(
     ([mode]) => {
-      window._playwrightSetMode(mode);
+      (window as any)._playwrightSetMode(mode);
```
justinkambic, line 298:
```suggestion
      // `_playwrightSetMode` is a private function
      (window as any)._playwrightSetMode(mode);
```

electron/execution.ts
```diff
@@ -293,21 +304,19 @@ async function onSetMode(mode) {
...
   } catch (e) {
     logger.error(e);
   }
 }
 
-function setupListeners() {
-  ipc.answerRenderer('record-journey', recordJourneys);
-  ipc.answerRenderer('run-journey', onTest);
+export default function setupListeners() {
+  ipc.answerRenderer<{ url: string }>('record-journey', recordJourneys);
```
justinkambic, line 316:
```suggestion
  ipc.answerRenderer<RecordJourneyOptions>('record-journey', recordJourneys);
```
Do you think it'd be worth it to make a named type here, and below?

electron/execution.ts
```diff
@@ -293,21 +304,19 @@ async function onSetMode(mode) {
...
+  ipc.answerRenderer<RunJourneyOptions>('run-journey', onTest);
   ipc.answerRenderer('save-file', onFileSave);
-  ipc.answerRenderer('actions-to-code', onTransformCode);
-  ipc.answerRenderer('set-mode', onSetMode);
-  ipc.answerRenderer('link-to-external', onLinkExternal);
+  ipc.answerRenderer<{ actions: Steps; isSuite: boolean }>('actions-to-code', onTransformCode);
```
justinkambic, line 319:
```suggestion
  ipc.answerRenderer<TransformCodeOptions>('actions-to-code', onTransformCode);
```

Same question about type.

src/common/types.ts
```diff
@@ -75,8 +75,9 @@ interface JourneyStartEvent {
 interface JourneyEndEvent {
```
justinkambic, line 75: If we are going to reference types on both the client and the server, it might be good to extract this types file to a top-level directory from which they both `import`, rather than having the `electron` code reference the `src` folder.

src/hooks/useSyntheticsTest.ts
```diff
@@ -106,3 +106,9 @@ export function useSyntheticsTest(steps: Steps): ITestContext {
     setResult,
   };
 }
+
+export type RunJourneyOptions = {
```
justinkambic, line 110: Why is this defined here?

justinkambic (changes requested):
This is great work and already seems to work fine.

I'd like to see a few things improve. I think we can tighten up our usage of `any` a bit.

Along the same line, I think we might want to change how we organize our types; we can move the types file to a top-level location from which all the code can import.

electron/execution.ts
kyungeunni: agreed 👍

electron/execution.ts
kyungeunni: It's either Buffer or string, but we know that Synthetics will emit string so I guess we can say it's string.

src/hooks/useSyntheticsTest.ts
kyungeunni: I _think_ the intention was to keep the type definition close to the function that requires this type. But it does look weird! I will move it under `shared`.

src/common/types.ts
kyungeunni: That is a great point!! I'm wondering if this common directory was supposed to be used by both client and server? if that's the case should we just move the whole directory to a top-level?

electron/execution.ts
kyungeunni: The type guard suggestion sounds fantastic 🌟 
However, IMHO, I'm not sure using `unknown` here would bring a big benefit here just because we need to cast the type of `parsed` one more time just before constructing a structured data from it(it seems to me like we will need to define a type for the data that lasts only temporarily).  I thought maybe we keep the parsed type as `Record<string, any>` and add the type checking with type guard on constructed data so that they 100% meet the type definition of `TestEvent`? WDYT?

electron/execution.ts
kyungeunni: Don't need this line. will delete

electron/execution.ts
vigneshshanmugam: Not sure why this is here in the first place 😆

tsconfig.json
```diff
@@ -9,13 +9,14 @@
     "strict": true,
     "forceConsistentCasingInFileNames": true,
     "noFallthroughCasesInSwitch": true,
-    "module": "esnext",
+    "module": "commonjs",
```
vigneshshanmugam, line 12: What we are doing with this is basically running CJS modules on the web (Its a nightmare). Use two separate build configs for frontend and Node.js. 
1. Create `tsconfig.base.json` - Use it as a common TS
2. Build Frontend app specific (react) - `tsconfig.frontend.json`
3. Build Node.js specific ones - `tsconfig.node.json`

src/common/types.ts
justinkambic: There is some "common" stuff that is specific to the frontend. A bunch of connector functions for interacting with the `ipc` channels that are only useful to the React code. I think it's alright to have a separate `common` directory at both levels, but if you think it's confusing we could maybe rename the frontend one to `shared` or some other thing. If we call it `shared`, let's rename the `shared.ts` to `index.ts` to avoid repeating the name. Then we could put the types in the root-level `common`.

electron/execution.ts
justinkambic: I'm fine with annotating it as `Buffer | string`, either option to me is better than explicitly naming it with `any` if we can avoid it.

electron/execution.ts
justinkambic: Sure, we can do that. The reason I preferred  `unknown` over `any` is that `unknown` will automatically flag any usage of potentially missing fields. But if we're going to expose all the code paths here to type guards anyway I suppose that's also an acceptable solution.

tsconfig.json
kyungeunni: Thanks for pointing it out 👍 One caveat is that create-react-app doesn't give us an option to specify tsconfig file path but only takes `tsconfig.json` (related ticket: https://github.com/facebook/create-react-app/issues/6023) so I needed to use tsconfig.json for frontend.

src/common/types.ts
kyungeunni: Thanks for clarifying it for me. I ended up creating `shared` at root directory and moved the type definition used by both electron and frontend to `types.d.ts`: https://github.com/elastic/synthetics-recorder/pull/205/commits/feeb4f5e6f4882fbf5b032f4caa99c75bf3f9eea WDYT?

electron/execution.ts
kyungeunni: Yeah I agree. I made the type definition for all non-primitive arguments 👍

electron/execution.ts
kyungeunni: Thanks, turns out this function expects only `StepEndEvent`(a subset of TestEvent) 👍 Thanks for spotting it!

electron/execution.ts
kyungeunni: Updated 👍

src/common/types.ts
justinkambic: The naming convention used in Kibana and a few other places is to call it `common`. Not super important but it'd be a small change.

electron/execution.ts
kyungeunni: I've added `if..else`s before returning the TestEvent objects as you suggested: https://github.com/elastic/synthetics-recorder/pull/205/commits/973d4a544647707e47d05f1167ef810cacf4366b

tsconfig.json
kyungeunni: I did push the changes. I specified the tsc to compile using `tsconfig.node.json` which compile the `electron` directory. AFAIK the react code is compiled by its own process using `tsconfig.json` (I specified to include only `src` dir)

electron/execution.ts
```diff
@@ -154,48 +167,59 @@ async function onTest(data, browserWindow) {
     });
   };
 
-  // returns TestEvent interface defined in common/types.ts
-  const constructEvent = parsed => {
+  const constructEvent = (parsed: Record<string, any>): TestEvent | null => {
     switch (parsed.type) {
```
justinkambic, line 171: I think if we did something like this we can improve this function a bit. Reduces the number of ternary/return statements and gives greater confidence and static typing for each return. Also pushes the verification logic into a separate function, which will make it easier to write tests for.

```typescript
  const isJourneyStart = (event: any): event is { journey: { name: string } } =>
    event.type === 'journey/start';

  const isStepEnd = (
    event: any
  ): event is {
    step: { duration: { us: number }; name: string; status: StepStatus };
    error?: Error;
  } => {
    return (
      event.type === 'step/end' &&
      ['succeeded', 'failed', 'skipped'].includes(event.step?.status) &&
      typeof event.step?.duration?.us === 'number'
    );
  };

  const isJourneyEnd = (
    event: any
  ): event is { journey: { name: string; status: 'succeeded' | 'failed' } } => {
    return event.type === 'journey/end' && ['succeeded', 'failed'].includes(event.journey?.status);
  };

  const constructEvent = (parsed: Record<string, any>): TestEvent => {
    if (isJourneyStart(parsed)) {
      return {
        event: 'journey/start',
        data: {
          name: parsed.journey.name,
        },
      };
    } else if (isStepEnd(parsed)) {
      return {
        event: 'step/end',
        data: {
          name: parsed.step.name,
          status: parsed.step.status,
          duration: Math.ceil(parsed.step.duration.us / 1000),
        },
      };
    } else if (isJourneyEnd(parsed)) {
      return {
        event: 'journey/end',
        data: {
          name: parsed.journey.name,
          status: parsed.journey.status,
        },
      };
    }
    return null;
  };
```

kyungeunni: Thanks, Justin - this is great! After reading this code, I understand the reason you suggested using `unknown` for the parsed(event in here) :) I really like how you used the type predicate here. I will also change the `any` to `unknown` as it works nicely. 🙏

justinkambic: Sounds fine - if we change `any` to `unknown` in the type guard functions, we might need to add some additional code to them to satisfy typescript as we are referencing fields on an `unknown` var now. I'm happy to pair on re-writing anything that doesn't drop right in.

.eslintrc.js
```diff
@@ -48,6 +48,7 @@ module.exports = {
       rules: {
         '@typescript-eslint/explicit-module-boundary-types': 0,
         '@typescript-eslint/no-non-null-assertion': 0,
+        '@typescript-eslint/no-explicit-any': 0,
```
kyungeunni, line 51: Hey @justinkambic I was wondering how do you feel about it?

justinkambic: I'm not crazy about it because I feel like our usage of `any` will be unrestricted and therefore more likely. But if we agree to not utilize `any` in situations where it's unnecessary but convenient, I suppose it's alright.

I usually prefer to disallow `any` and cover it with `@ts-expect-error We're using any because....`.

justinkambic (approved):
LGTM!

electron/execution.ts
kyungeunni: Thanks, I personally think it should be fine with `any` in this case

PR #206, opened 2022-04-14
fix: Header layout when paused

### LINKED ISSUE #202: [Bug] Header layout is not maintained when Recorder is Paused
### Bug summary

When the recorder is in pause mode, the header's layout ceases to fill the entire area.

### Recorder Version

alpha-3

### Steps to reproduce

1. Start recording
2. Pause
3. Observe the layout change

### Expected behavior

The recorder's header should use the same layout regardless of recording status.

### Additional information

_No response_

justinkambic:
## Summary

Resolves #202.

Fixes header layout when Paused.

## Implementation details

One-line change.

### Before

image

### After

image

### DIFF
--- a/src/components/Header/HeaderControls.tsx
+++ b/src/components/Header/HeaderControls.tsx
@@ -110,7 +110,12 @@ export function HeaderControls({ setIsCodeFlyoutVisible }: IHeaderControls) {
           </ControlButton>
         </EuiFlexItem>
       )}
-      <EuiFlexItem grow={recordingStatus === RecordingStatus.Recording}>
+      <EuiFlexItem
+        grow={
+          recordingStatus === RecordingStatus.Recording ||
+          recordingStatus === RecordingStatus.Paused
+        }
+      >
         <RecordingStatusIndicator status={recordingStatus} />
       </EuiFlexItem>
       <EuiFlexItem grow={false}>


src/components/Header/HeaderControls.tsx
```diff
@@ -110,7 +110,12 @@ export function HeaderControls({ setIsCodeFlyoutVisible }: IHeaderControls) {
           </ControlButton>
         </EuiFlexItem>
       )}
-      <EuiFlexItem grow={recordingStatus === RecordingStatus.Recording}>
+      <EuiFlexItem
+        grow={
+          recordingStatus === RecordingStatus.Recording ||
+          recordingStatus === RecordingStatus.Paused
+        }
+      >
```
kyungeunni, line 118: Would this work? (I've tested locally but not sure if I missed some reasons behind it)
```suggestion
      <EuiFlexItem grow>
```

kyungeunni, line 118: Would this work? (I've tested locally but not sure if I missed some reasons behind it)
```suggestion
      <EuiFlexItem grow>
```

justinkambic: No this won't work because we want it to `grow` only when status is `Recording` or `Paused`. When we are `NotRecording`, we want the URL input to expand to fill the rest of the space in this row. If we `grow` always, the input will not be able to fill the entire space:

image

vs. if we use the code as written:

image

justinkambic: Sometimes they're a necessary evil 😆

PR #208, opened 2022-04-14
fix: Reset default values when action/assertion edit view closes

### LINKED ISSUE #201: [Bug] Status indicator for Assertion is missing and unsaved state persists for actions
### Bug summary

Found a couple of bugs while testing #65

- Unlike other actions, assert action doesn't have a status indicator 

Screenshot 2022-04-13 at 18 35 17

- When editing an action, I could choose to cancel the changes I made. If I go back to edit the action again, the previous state(right before canceling) persists and it is different from the real value stored within the action.
 [Image from Gyazo](https://gyazo.com/969242d76cdd84e90ae1449cafabf5ab)

### Recorder Version

0.0.1-alpha.3

### Steps to reproduce

- Record a script, hover on action, and click _Add assertion_ in the control menu. See the status indicator is not visible

- Record a script, hover on action, and click _Edit action_ in the control menu. Change the value in any of the input elements then click _cancel_ button. Click _Edit action_ again for the same action and you will see the unsaved change persists in the input element rather than the actual value shown in the action summary.

### Expected behavior

- Status indicator should be shown to all the actions including `assert` action
- Unsaved changes shouldn't persist when editing actions.

### Additional information

_No response_

justinkambic:
## Summary

Resolves #201.

## Implementation details

Conditionally renders the child for `ActionElement`, which will cause all local state to be destroyed when the element is closed.

The side effect of this change is that the default values for the edit forms will reset between close/open events.

Open an assertion and an action element, modify some values in the forms, then close and reopen the elements. Your ephemeral changes should be lost.

### DIFF
--- a/e2e/tests/assertionInfo.test.ts
+++ b/e2e/tests/assertionInfo.test.ts
@@ -41,9 +41,6 @@ describe('Assertion Info Popover', () => {
       'id=action-element-0-0',
       'text=Add assertion'
     );
-    await electronWindow.click('id=action-element-0-1');
-    await electronWindow.hover('id=action-element-0-1');
-    await electronWindow.click(`[aria-label="Begin editing this action"] >> nth=1`);
     await electronWindow.click(
       `[aria-label="Shows a popover with more information about Playwright assertions."]`
     );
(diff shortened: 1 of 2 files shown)


kyungeunni:
Tested locally, LGTM!! thank you ⭐

kyungeunni (approved):

PR #209, opened 2022-04-14
chore: Simplify generator unit test

### LINKED ISSUE #193: [Other] Simplify generator unit tests
### Task summary

We have unit tests covering the behavior of the [`generator`](https://github.com/elastic/synthetics-recorder/blob/main/src/helpers/generator.test.ts) code in the recorder; these are quite bloated with boilerplate, and we can greatly reduce the volume of the code they contain using the [`createAction`](https://github.com/elastic/synthetics-recorder/blob/main/src/helpers/test/createAction.ts) helpers we have recently introduced.

### Linked resources

_No response_

### Additional information

_No response_

justinkambic:
## Summary

Resolves #193.

## Implementation details

Adds some new unit test creator functions that accept override objects to specify custom field values.

Reduces the complexity of the generator tests by a bit. Can probably be further reduced.

Unit tests should all continue passing.

### DIFF
--- a/src/helpers/test/createAction.ts
+++ b/src/helpers/test/createAction.ts
@@ -30,13 +30,16 @@ type ActionWithName = Partial<Action> & { name: string };
 type CreateStepActionOverride = Partial<Omit<ActionContext, 'action'>> & { action: ActionWithName };
 type ActionContextOverride = Partial<Omit<ActionContext, 'action'>> & { action?: ActionOverride };
 
-export const createAction = (name: string, overrides?: ActionContextOverride): ActionContext =>
-  overrides
+export const createAction = (name: string, overrides?: ActionContextOverride): ActionContext => {
+  const baseAction = {
+    frameUrl: 'https://www.elastic.co',
+    isMainFrame: true,
+    committed: true,
+    pageAlias: 'pageAlias',
+  };
+  return overrides
     ? {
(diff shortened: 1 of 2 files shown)


src/helpers/test/createAction.ts
```diff
@@ -22,22 +22,54 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
...
+        action: {
+          name,
+          signals: [],
+        },
+        frameUrl: 'https://www.elastic.co',
+        isMainFrame: true,
+        committed: true,
+        pageAlias: 'pageAlias',
+      };
+
```
lucasfcosta, line 58: NIT:

```suggestion
export const createAction = (name: string, overrides?: ActionContextOverride): ActionContext => {
  const baseAction = {
        frameUrl: 'https://www.elastic.co',
        isMainFrame: true,
        committed: true,
        pageAlias: 'pageAlias',
  };
  
  return overrides
    ? {
        ...baseAction,
        ...overrides,
        action: {
          name,
          signals: [],
          url: 'https://www.elastic.co',
          ...overrides.action,
        },
      }
    : {
        ...baseAction,
        action: {
          name,
          signals: [],
        },
      };
}
```

lucasfcosta (approved):
Much nicer, thank you!

PR #211, opened 2022-04-20
fix: Hoist page objects in generated code

### LINKED ISSUE #185: [Chore] Migrate generator code out of `@elastic/synthetics` to recorder
### Task summary

The code responsible for generating step/action code today lives in `@elastic/synthetics`. The `synthetics` repo needs to update its Playwright version to resolve user issues, and because we are using a doctored fork of Playwright in the recorder, we will need to diverge for a while. This means we cannot use future releases of `synthetics` for the time being.

~As part of this temporary divergence, we will move the code that currently lives in `synthetics` over to the recorder so we are able to maintain and change it as needed without having to update the `synthetics` dependency version.~

We are going to copy this generator code to the Recorder and modify it as needed for now. We will have to reconcile the version living in Synthetics with our downstream modifications at some point in the future.

### Linked resources

https://github.com/elastic/synthetics/blob/main/src/formatter/javascript.ts

### Additional information

At the end of this change, the types for `Steps`, `Step`, `ActionInContext`, `Action`, should all live in the recorder, as well as the code necessary to generate JavaScript code from the actions users create using the Recorder.

### LINKED ISSUE #195: [Bug] Hoist page objects to journey level in the synthetics generator
### Bug summary

When multiple page objects are created in the recorder session either by 
   - clicking on links that moves from `a.com` to `b.com`
   - Opening popups from `a.com` to `b.com`

Playwright creates page objects for each of those pages, The Synthetics JS generator then creates the necessary code to express those sessions. Take an example below
```js
step('Go to https://vigneshh.in/', async () => {
  await page.goto('https://vigneshh.in/');
});
step('Click text=Tailor', async () => {
  const [page1] = await Promise.all([
    page.waitForEvent('popup'),
    page.click('text=Tailor')
  ]);
  await page1.click('text=Packages 0');
  expect(page1.url()).toBe('https://github.com/orgs/zalando/packages?repo_name=tailor');
  await page1.close();
});
step('Click text=Babel Minify', async () => {
  const [page2] = await Promise.all([
    page.waitForEvent('popup'),
    page.click('text=Babel Minify')
  ]);
  await page2.close();
});
```

The problem here is we want the variable declaration for `page1` and `page2` and their `close` method to be on their dedicated steps or grouped on a single step. This issue was mitigated before as we controlled the time when steps gets created which was only on `clicks` that triggered new navigations. 

But with the introduction of Step divider UI, Its way easier for our users to create steps for each action and run in to the `variable not being defined` issue. Creating step before `page1 close` would yield this code. 

```js
step('Go to https://vigneshh.in/', async () => {
  await page.goto('https://vigneshh.in/');
});
step('Click text=Tailor', async () => {
  const [page1] = await Promise.all([
    page.waitForEvent('popup'),
    page.click('text=Tailor')
  ]);
  await page1.click('text=Packages 0');
  expect(page1.url()).toBe('https://github.com/orgs/zalando/packages?repo_name=tailor');
});
step('Close page', async () => {
  await page1.close();
});
step('Click text=Babel Minify', async () => {
  const [page2] = await Promise.all([
    page.waitForEvent('popup'),
    page.click('text=Babel Minify')
  ]);
  await page2.close();
});

```

### Recorder Version

main branch

### Steps to reproduce

1. Start the recorder and type any page url.
2. Taking `www.vigneshh.in` for now as the test was performed on that page.
3. Click on `a` links that opens new page objects.
4. Try creating step dividers and run `Test` button.

Screen Shot 2022-04-12 at 12 46 27 PM

### Expected behavior

+ Synthetics tests should run without any `JS errors`. 

### Additional information

The easiest way to solve this would be to hoist the page variables before the step function declaration and adds guards when functions are invoked on them.

```js
let page1;
step('Click text=Tailor', async () => {
  page1 = await Promise.all([
    page.waitForEvent('popup'),
    page.click('text=Tailor')
  ]);
  await page1.click('text=Packages 0');
  expect(page1.url()).toBe('https://github.com/orgs/zalando/packages?repo_name=tailor');
});
step('Close page', async () => {
  await page1.close();
});

```

Another approaches, Google chrome recorder tries to create separate blocks for each of the step, but requires you to be on the recorder page to work correctly. 

```js
{
    const targetPage = page;
    const element = await waitForSelectors([["aria/Tailor"],["body > div > div.profile > div.content.left > p:nth-child(5) > a:nth-child(1)"]], targetPage, { timeout, visible: true });
    await scrollIntoViewIfNeeded(element, timeout);
    await element.click({ offset: { x: 32.0390625, y: 8.90625} });
}
{
    const target = await browser.waitForTarget(t => t.url() === "https://github.com/zalando/tailor", { timeout });
    const targetPage = await target.page();
    targetPage.setDefaultTimeout(timeout);
    await targetPage.keyboard.down("Meta");
}
```

Open to new ideas as well, If anyone wants to tackle, I am more than happy to pair on this one.

justinkambic:
## Summary

Resolves #195.

Resolves #185.

Depends on #205.

## Implementation details

Copies generator code from synthetics to recorder. This is done so we can modify it to fix a defect that can cause a script to reference out-of-scope vars.

You can see steps on how to re-create this problem in the linked issue. Ensure that you cannot create code that runs into this problem, and that the code we generate continues to work as designed.

Some of the code in this PR is copied directly from Synthetics, other bits are new to this change. Particular code areas to check out:

- the [function](https://github.com/elastic/synthetics-recorder/pull/211/files#diff-7c0276f32ead59b6dfb141ee9ee1426f428d8528531be9ec9f9383cae0bec593R353) that identifies the vars to hoist
- [assignment](https://github.com/elastic/synthetics-recorder/pull/211/files#diff-7c0276f32ead59b6dfb141ee9ee1426f428d8528531be9ec9f9383cae0bec593R311) of the generator's vars to hoist
- [updated conditions](https://github.com/elastic/synthetics-recorder/pull/211/files#diff-7c0276f32ead59b6dfb141ee9ee1426f428d8528531be9ec9f9383cae0bec593R204) in `generateAction` to account for whether a var is hoisted during its definition or not
- [updated logic](https://github.com/elastic/synthetics-recorder/pull/211/files#diff-7c0276f32ead59b6dfb141ee9ee1426f428d8528531be9ec9f9383cae0bec593R235) for finishing a promise all emit procedure


kyungeunni (commented):
Great work Justin 👍  I noticed a couple of things.

1. The error message with the code preview is not visible on the flyout when test failed:
image

2. It is not reliably reproducible, but I've encountered this case when deleting and adding some steps, the test would fail and this error was printed in the console. I will try to find a way to reproduce it.
image

electron/syntheticsGenerator.ts
```diff
@@ -0,0 +1,365 @@
...
+  getDefaultFormatterOffset() {
+    return this.isSuite ? 2 : 0;
+  }
+
+  /**
+   * We need to hoist any page or popup alias that appears in more than one step.
+   * @param steps the step IR to evaluate
+   * @returns an array that contains the names of all variables that need to be hoisted
+   */
+  findVarsToHoist(steps: Steps): string[] {
```
kyungeunni, line 348: After seeing 3-level nested loops I was hoping to find a bit simpler way, but I couldn't come up with anything better so I will leave it to it. :)

justinkambic: The complexity is unfortunately tied to the data structure. We could create some recursive function instead if you prefer; would reduce nesting but increase the logical acuity needed to trace the code.

electron/syntheticsGenerator.ts
```diff
@@ -0,0 +1,363 @@
...
+  throw new Error('Invalid escape char');
+}
+
+type Formattable = string | string[] | Record<string, unknown>;
+
+function isFormattable(value: unknown): value is Formattable {
+  return (
+    typeof value === 'string' ||
+    (Array.isArray(value) &&
+      value.reduce<boolean>((prev, cur) => prev && typeof cur === 'string', true)) ||
```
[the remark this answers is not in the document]
justinkambic: 271566b93b3dc41353bed5109c87c10fc08dea5d

justinkambic:
> The error message with the code preview is not visible on the flyout when test failed

This happens on `main` as well. I've created https://github.com/elastic/synthetics-recorder/issues/232 to track this.

electron/syntheticsGenerator.ts
```diff
@@ -0,0 +1,363 @@
...
+    typeof value === 'object'
+  );
+}
+
+function formatObject(value: string | string[] | Record<string, unknown>, indent = '  '): string {
```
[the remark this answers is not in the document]
justinkambic: https://github.com/elastic/synthetics-recorder/pull/211/commits/0970e46c10da6062951585319a219001b3ffc4cc

justinkambic:
> It is not reliably reproducible, but I've encountered this case when deleting and adding some steps, the test would fail and this error was printed in the console.

Add a `console.log(steps)` to `App.tsx` and see if you can repro, then send me the data in a gist or something so I can reproduce/fix. I'll investigate myself as well.

electron/syntheticsGenerator.ts
kyungeunni: Yes, I agree that this is due to the data structure. It seems to be best to leave it like that and avoid additional effort just to avoid the nested loops. Thanks for sharing your thoughts.

electron/syntheticsGenerator.ts
```diff
@@ -0,0 +1,364 @@
...
+      formatter.add(`let ${varName};`);
+    }
+    return formatter.format();
+  }
+
+  isVarHoisted(varName: string) {
+    return this.varsToHoist.indexOf(varName) >= 0;
+  }
+
+  getDefaultFormatterOffset() {
```
vigneshshanmugam, line 338: nit: getDefaultOffset, formatter is wordy as we are already in the formatter.

electron/syntheticsGenerator.ts
```diff
@@ -0,0 +1,364 @@
...
+      // const [popup1] = await Promise.all([]).
+      let leftHandSide = '';
+      if (signals.popup && signals.popup.popupAlias && !isVarHoisted) {
+        leftHandSide = `const [${signals.popup.popupAlias}] = `;
+      } else if (isVarHoisted) {
+        leftHandSide = `${signals.popup?.popupAlias} = `;
+      } else if (signals.download) {
+        leftHandSide = `const [download] = `;
+      }
+      const rightHandSide = isVarHoisted ? `(await Promise.all([` : `await Promise.all([`;
```
vigneshshanmugam, line 207: Any reason why we are doing this instead of array assignment? 
```ts
[page] = await Promise.all([]) 

// vs

page = (await Promise.all([]))[0]
```

I would keep out formatter close to the PW's to keep the changes minimal and move if we support the pw-test-runner in the future. Also I feel the former is cleaner.

electron/syntheticsGenerator.ts
```diff
@@ -0,0 +1,364 @@
...
+      ? `${pageAlias}.frame(${formatObject({
+          name: actionInContext.frameName,
+        })})`
+      : `${pageAlias}.frame(${formatObject({
+          url: actionInContext.frameUrl,
+        })})`;
+
+    const signals = toSignalMap(action);
+
+    const isVarHoisted = signals.popup?.popupAlias && this.isVarHoisted(signals.popup.popupAlias);
```
vigneshshanmugam, line 187: move this block inside `emitPromiseAll` - not sure if it's relevant for other parts of the code.

electron/syntheticsGenerator.ts
```diff
@@ -0,0 +1,364 @@
...
+  }
+
+  /**
+   * We need to hoist any page or popup alias that appears in more than one step.
+   * @param steps the step IR to evaluate
+   * @returns an array that contains the names of all variables that need to be hoisted
+   */
+  findVarsToHoist(steps: Steps): string[] {
+    const pageAliasCounts = steps.reduce((acc, step) => {
+      const aliases = new Set<string>();
```
vigneshshanmugam, line 349: This algorithm can be simplified by having this two thoughts
- No need to distinguish between steps, since no of alias is proportional to the number of pages created by the user. 
- Hoist the page alias no matter what even there is only one step.

Having that assumptions, the code would look like this
- Having a single set that maintains the page alias
- Go over the popups and aliases and add it to the Set
- Produce the generated script except for the `page` object which is maintained by Synthetics.

electron/syntheticsGenerator.ts
justinkambic: It's one of those things I changed quickly as I was focusing on something else; I needed to get this var into a higher scope and it seems I went too high.

electron/syntheticsGenerator.ts
justinkambic: Ok, that is simpler, thanks for the suggestion. My original motivation was to avoid hoisting unless it was necessary, this way we'd keep non-glitchy scripts as close to the original format as possible. But this is indeed a simpler approach and if you're ok with it I'm fine with it as well. Thanks for the detailed comment!

electron/syntheticsGenerator.ts
justinkambic: https://github.com/elastic/synthetics-recorder/pull/211/commits/de8dfafdee1899f0c71b20746e46b0a92932ef8f

electron/syntheticsGenerator.ts
justinkambic: https://github.com/elastic/synthetics-recorder/pull/211/commits/52ab47b985240618ef9cee45f2bc855cca750ee7

electron/syntheticsGenerator.ts
justinkambic: https://github.com/elastic/synthetics-recorder/pull/211/commits/638d8bd1b00943faa45c9139f3a41a352b072d1e

electron/syntheticsGenerator.ts
justinkambic: @vigneshshanmugam [suggested](https://github.com/elastic/synthetics-recorder/pull/211#discussion_r861388311) a few ways to improve it below, I've added those.

electron/__snapshots__/syntheticsGenerator.test.ts.snap
```diff
@@ -0,0 +1,110 @@
...
+    page.waitForEvent('popup'),
+    page.click('text=Tailor')
+  ]);
+  await page1.click('text=Packages 0');
+  expect(page1.url()).toBe('https://github.com/orgs/zalando/packages?repo_name=tailor');
+  await page1.close();
+  [page2] = await Promise.all([
+    page.waitForEvent('popup'),
+    page.click('text=Babel Minify')
+  ]);
```
kyungeunni, line 47: It is quite clear that page1 doesn't exist anymore so it is obvious to use `page` in line 45 and 46. But out of curiosity, what happens when page1 isn't closed yet - then do we know which one from `page` or `page1` need to 
 be used in lines 45 and 46 ?

kyungeunni (approved):
I appreciate your extra work on findVarsToHoist function!! looking great!! ✨

electron/__snapshots__/syntheticsGenerator.test.ts.snap
vigneshshanmugam: Pages are the context in PW, It does not matter if the page 1 window is kept open or closed, If the
- new window is opened from page1, then page2 would be waiting for popup event from `page1`. 
- If the window is opened from first page `page`, then page2 would be waiting for popup from `page` . 

Hope this clears up.

PR #212, opened 2022-04-20
fix: use action.url instead of frameUrl in header

### LINKED ISSUE #200: [Bug] Fix flaky e2e test

### LINKED ISSUE #210: [Bug] Action header shows stale value after editing url
### Bug summary

When updating the URL value in navigate action, the action header still shows the previous/stale value. But when I expand the edit panel again or click the Export button, it shows the correct value.

[Image from Gyazo](https://gyazo.com/8617b07c87d1db3c393b93646c57b9b3)

### Recorder Version

0.0.1-alpha.3

### Steps to reproduce

1. Record a script
2. Edit navigate action, input different URL and click save button
3. After the edit panel is collapsed, Observe the value in the Action header is still the old value

### Expected behavior

The action header should display the correct value

### Additional information

_No response_

kyungeunni:
fixes #210 
(possibly fixes #200)

## Summary

It shows the correct URL in Action header after updating the url value

[Image from Gyazo](https://gyazo.com/7f3be4677f83ef4f38f8df152deffa98)
## Implementation details
Use actionInContext.action.url instead of actionInContext.frameUrl

Update url in navigate action and notice that the url in action header is also updated to the correct value
Also the test will expect for the element header to contain the new value

### DIFF
--- a/e2e/tests/actionValues.test.ts
+++ b/e2e/tests/actionValues.test.ts
@@ -54,7 +54,7 @@ async function editAssertion(electronWindow: Page) {
   await electronWindow.fill(`[aria-label="Assertion selector"]`, ASSERTION_SELECTOR);
   await electronWindow.fill(`[aria-label="Assertion value"]`, ASSERTION_VALUE);
   await electronWindow.click(`[data-test-subj="save-0-1"]`);
-  // make sure state has been updated by checking the ation header's content
+  // make sure state has been updated by checking the action header's content
   await electronWindow.waitForSelector(`#action-element-0-1 >> div:has-text("Inner Text")`, {
     timeout: 2000,
   });
--- a/e2e/tests/actionValues.test.ts
+++ b/e2e/tests/actionValues.test.ts
@@ -54,6 +54,10 @@ async function editAssertion(electronWindow: Page) {
   await electronWindow.fill(`[aria-label="Assertion selector"]`, ASSERTION_SELECTOR);
   await electronWindow.fill(`[aria-label="Assertion value"]`, ASSERTION_VALUE);
   await electronWindow.click(`[data-test-subj="save-0-1"]`);
+  // make sure state has been updated by checking the ation header's content
+  await electronWindow.waitForSelector(`#action-element-0-1 >> div:has-text("Inner Text")`, {
+    timeout: 2000,
+  });
 }
 
 async function editAction(electronWindow: Page) {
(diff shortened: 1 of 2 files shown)


justinkambic (approved):
LGTM! Closed my other fix PR in favor of this one, great job.

Maybe run a few more `/test` comments through CI to see if it flakes at all. If it's the same issue we are trying to solve it should only take 2-3 times for it to fail.

PR #213, opened 2022-04-21
fix: Clear steps as part of `startOver` procedure

### LINKED ISSUE #180: [Bug] Previous script's state persists when re-recording
### Bug summary

When re-recording the script, I expect the previous script to be deleted after I click _Delete and start over_ button, but the state persists until the browser is launched. It gets cleared eventually but I expect the clean state when waiting for the browser to start.

[Image from Gyazo](https://gyazo.com/aa39c3ceb84f84dcf2afd8734f73cfbd)

### Recorder Version

0.0.1-alpha.2

### Steps to reproduce

1. Record a script 
2. Click `Start over` button. When confirm modal shows up, click _Delete and start over_  button.
3. Observe the previous script stays there while it changed to _recording_ state.
4. When browser is launched, go back to Recorder and see now the previous script is deleted.

### Expected behavior

Stale script should be deleted as soon as clicking Delete and start over button

### Additional information

_No response_

justinkambic:
## Summary

Resolves #180.

## Implementation details

Add a call to `setSteps([])` to clear the local step data just before we start a new recording session.

1. Record a journey.
2. Click "Start Over" button.
3. Observe that the old step data is gone.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -73,7 +73,8 @@ export default function App() {
     ipc,
     url,
     steps.length,
-    syntheticsTestUtils.setResult
+    syntheticsTestUtils.setResult,
+    setSteps
   );
   const { isStartOverModalVisible, setIsStartOverModalVisible, startOver } = recordingContextUtils;
   const dragAndDropContext = useDragAndDropContext();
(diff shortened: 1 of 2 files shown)


lucasfcosta (approved):
LGTM!

I've also tested this PR E2E and the steps I had previously recorded have been cleared immediately once I clicked the start over button (even before the browser actually opened).

clear_steps

PR #216, opened 2022-04-21
feat: Add info icon and tooltip to export as script checkbox

### LINKED ISSUE #166: [Enhancement] Add a tooltip to the suite checkbox
### Feature summary

We should add a tooltip to the suite checkbox - export a script flow

Screenshot 2022-03-01 at 15 18 11

Screenshot 2022-03-01 at 15 41 19

For users that are not familiar with the suite concept, this can give more context. 

### Linked resources

_No response_

### Additional information

_No response_

justinkambic:
## Summary

Resolves #166.

## Implementation details

image

Adds an info icon with a tooltip to add additional context to the `export as script` checkbox.

Record a journey, click the export button, and hover over the `iInCircle` icon.

### DIFF
--- a/src/components/ExportScriptFlyout/Body.tsx
+++ b/src/components/ExportScriptFlyout/Body.tsx
@@ -22,7 +22,16 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 */
 
-import { EuiCheckbox, EuiCodeBlock, EuiFlyoutBody, EuiSpacer } from '@elastic/eui';
+import {
+  EuiCheckbox,
+  EuiCodeBlock,
+  EuiFlexGroup,
+  EuiFlexItem,
+  EuiFlyoutBody,
+  EuiIcon,
+  EuiSpacer,
+  EuiToolTip,
+} from '@elastic/eui';
 import React from 'react';
 import type { Setter } from '../../common/types';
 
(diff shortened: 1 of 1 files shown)


kyungeunni:
Probably worth checking with @paulb-elastic and @drewpost about the wording?

lucasfcosta (approved):
LGTM, also E2E verified.

tooltip

paulb-elastic:
The description doesn’t make it too clear how ticking/unticking makes it inline/suite.  I’m not massively happy with the following (would be good to get @colleenmcginnis's input too), but maybe something like this:

>By default, the script code will be exported in a format that can be pasted as an inline script when adding a monitor, or being dragged into the “Script recorder” option. If you are going to add the monitor from a ZIP URL (which will run as a suite), then select to export it as a suite.

Note that we may want to change this wording when we get the new Synthetics UI (with new names for inline etc.) and also the PUSH command to replace the ZIP URL option that's in the current Synthetics integration.

colleenmcginnis:
We could try to make this clearer with words, but I'm wondering if a checkbox is the right UI element to communicate these options. Maybe this is more a question for @hbharding, but if there are two export options (inline and suite) would something like a [toggle button](https://elastic.github.io/eui/#/navigation/button#toggle-buttons) be more appropriate? 

image

hbharding:
I think a toggle button or switch component could work well here. Since changing the option causes a visual change (the code snippet below changes, if I remember correctly), a toggle or light switch feels more appropriate.

While I don't love the names "inline" and "suite", I think a two button toggle is more clear since we would have to spell out "Inline" (the default option). But, if we don't need to expose this terminology, a binary switch toggle "Export as suite" with tooltip could work too.

justinkambic:
So, based on offline discussion I think we are actually closer to closing this PR and canceling the original feature request. There's been a lot of discussion about removing the ability to export as suite altogether from the recorder. I think we likely won't merge this despite it being approved.

PR #218, opened 2022-04-21
fix: Disable record journey button when URL input is empty

### LINKED ISSUE #126: [Enhancement] Preventing users from starting a journey without an initial URL
### Feature summary

The users should not be able to start recording a journey without an URL, otherwise, we'll record a step in which they're taken to Chromium's `about:blank` page.

We should only enable the record button after the user enters a `URL` into the recorder's URL `input`.

I assume that no users will want their scripts to have a step in which the browser goes to an `about:blank` page. It will be just extra work for them to remove it.

Furthermore, IMO this enhancement makes the user more attentive to the URL field as they will have to enter a URL to start recording.

### Linked resources

_No response_

### Additional information

We still need to check whether we really want to implement this "enhancement". I don't think it's a big deal, but it does raise the user's attention to using the initial URL input properly.

justinkambic:
## Summary

Resolves #126.

## Implementation details

Disables the start recording button if the URL field is empty.

image

1. Launch the recorder and try to start a journey without a URL.
1. Enter a URL and start recording a journey to ensure the original functionality remains intact.

### DIFF
--- a/src/components/Header/HeaderControls.tsx
+++ b/src/components/Header/HeaderControls.tsx
@@ -87,7 +87,7 @@ export function HeaderControls({ setIsCodeFlyoutVisible }: IHeaderControls) {
         <ControlButton
           aria-label={getPlayControlCopy(recordingStatus, steps.length)}
           color="primary"
-          isDisabled={isTestInProgress}
+          isDisabled={isTestInProgress || !url}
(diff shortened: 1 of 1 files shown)


kyungeunni (approved):
LGTM :shipit:

PR #220, opened 2022-04-25
chore: Eject create react app

### LINKED ISSUE #190: [Other] Explore possibility of removing `create-react-app`
### Task summary

Today we rely on `create-react-app` for the automatic bundling and serving of our frontend. This is really convenient because it eliminates a lot of maintenance and setup tasks that we would otherwise need to implement manually.

This also has drawbacks.

1. We have no control over updating dependencies that are out of date with security concerns. There are several and we are already on the latest version of the package.
2. We have limited control over how its internal webpack is working, and if we wanted to explore optimization of our bundle size or other build-related tasks, we'd need to do this without using `create-react-app`.
3. We have tangential dependencies like `wait-on` that we need to use for concurrency when scripting our dev mode. It might be possible to eliminate those as well with a custom solution.
4. Expanding typescript support to the backend is complicated because we need multiple configs and it's much more difficult to work around `create-react-app`'s internal configurations.

For these reasons I think it might be worth exploring the elimination of `create-react-app` from the codebase. The tradeoff to reduced dependency load and greater control over the build process is that we would have additional setup code of our own to maintain/update as the application changes and grows over time.

### Linked resources

_No response_

### Additional information

_No response_

justinkambic:
## Summary

Resolves #190.

This contains the output of running `create-react-app`'s `eject` script. This extricates `create-react-app` as a dependency, and moves all of its scripting and other configuration to the top-level of the repo.

This is a one-way destructive action, so if we merge this we cannot go back to `create-react-app`.

## Implementation details

This is a draft PR that may never end up being merged. It's done to better inform our decision to either remove `create-react-app` or to rely on a custom fork.

Will update if looking to merge


justinkambic:
Closing this in favor of using a feature branch to allow for incremental reduction of the maintenance burden before we merge to `main`.

PR #226, opened 2022-04-26
feat: Add status indicator to assertion elements

### LINKED ISSUE #222: [Enhancement] Display assertions as actions in UI
### Feature summary

Today, the script recorder shows assertions without a status indicator, differentiating them from Actions. We feel that, since Assertions are a specialized version of an Action, we should display them in the same manner as actions.
Untitled

We feel that the UI as it sits does not represent what is actually happening in the generated script.

### Linked resources

This is a continuation of the work defined in https://github.com/elastic/synthetics-recorder/issues/65.

### Additional information

_No response_

justinkambic:
## Summary

Resolves #222.

Adds a status indicator to assertion elements as well as action elements.

### Before

image

### After

image

1. Create an assertion.
1. See the status indicator.

### Other things to check

Ensure the status color changes after a test run

### DIFF
--- a/src/components/ActionElement/ActionElement.tsx
+++ b/src/components/ActionElement/ActionElement.tsx
@@ -122,7 +122,7 @@ function ActionComponent({
         />
       </EuiFlexItem>
       <EuiFlexItem grow={false}>
-        {!isAssertion && <ActionStatusIndicator showRect={isLast} status={testStatus} />}
+        <ActionStatusIndicator showRect={isLast} status={testStatus} />
       </EuiFlexItem>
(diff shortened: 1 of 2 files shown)


lucasfcosta (approved):
LGTM! Also E2E confirmed that:

1. Status indicators appear when adding the assertion
2. Tests show the correct status for the assertion


Screenshot 2022-04-27 at 09 24 20
Screenshot 2022-04-27 at 09 24 01

Screenshot 2022-04-27 at 09 24 30



Btw as a note, the name of the action itself within the step is duplicated but I think we have already discussed that in another issue and will be handled separately in the future.

PR #227, opened 2022-04-27
ci: refactor the docker build stage

v1v:
## Summary

Move away from CI specific dockerfile to a local development approach dockerfile based

## Implementation details

1) The CI will be the one generating the required docker image, if needed, and pushed to the docker-registry.
2) The CI will use the cached docker image to run the tests.
3) The docker container will mount the workspace instead copying the workspace to a different location, therefore everyone can use their workspace locally.
4) Run the `Build` stage before the `Test`, 
  a) current docker container runs as root so any generated files will be owned by a different user, so if needed to run `Test` before `Build` then it should be fixed the permissions.
5) `/test all` github command allows to force the Docker build.
6) PRs without changing the Dockerfile, will consume the docker image from the docker registry, the one tagging their target branch, in other words, if PR-1234 targeting `main`, then it will pull  the docker image `docker.elastic.co/observability-ci/synthetics-recorder:main`.

## Why

1. The current implementation can run in the CI controller.
2. It always builds the docker image, therefore increases the build time.
3. It cannot be reproduced locally but in the CI.
4. It uses absolute path and produce issues with the `junit` step.

Currently it takes `5 minutes` to prepare the docker image

image

```bash
$ .ci/scripts/docker-build.sh
$ .ci/scripts/run-test-in-docker.sh
```

### If there are changes in the Dockefiles then

From `15 minutes` to `12 minutes`. This PR is the one that shows those build times

| PR | Baseline |
| ---| ---------|
| image | image |

### If no changes in the Dockerfiles then

Potentially, I foresee about `10 minutes` ~ `30 %` faster. 

https://github.com/elastic/synthetics-recorder/pull/228 is the one to test and validate the above assumption.

Update, it took nearly `11 minutes` ~ `27 %` faster

image

## Issue

https://github.com/elastic/synthetics-recorder/pull/110


kuisathaverat (approved):

.ci/scripts/run-test.sh
```diff
@@ -8,6 +8,7 @@ echo $USER
 echo $USERNAME
 
 source $NVM_DIR/nvm.sh
-cd $DOCKER_BASE_DIR
 nvm use
+## Let's use install rather than ci to help with a faster execution
+npm install
```
lucasfcosta, line 13: I think this could be dangerous because `npm install` will always use the latest version of deps which aren't fixed with their exact versions in `package.json`, thus causing the tests to fail if a dependency has breaking changes which haven't properly bumped `major` or if we had tests which relied on a particular buggy behaviour which has changed.

IMO we should still use `npm ci` even if that means longer build times. It also ensures that the code we're testing is the exact code we'll ship.

More details on the behaviour of `ci` vs. `install` here: https://lucasfcosta.com/2020/10/17/lockfile-guide.html

v1v: e8776ae is the proposal:
1) In the CI, it relies in the first stage about `npm ci` since the docker container mounts the workspace as volume
2) Locally, it's possible to customise the behaviour with a new argument:

```bash
$ .ci/scripts/run-test-in-docker.sh install
$ .ci/scripts/run-test-in-docker.sh ci
```

lucasfcosta: Looks great, thank you! I'm very happy that we're mounting the workspace as a volume and thus can reuse deps rather than reinstalling everything. Thanks for that.

lucasfcosta (approved):

PR #229, opened 2022-04-27
chore: Eject create react app

### LINKED ISSUE #190: [Other] Explore possibility of removing `create-react-app`

justinkambic:
## Summary

Resolves #190.

This PR tracks the WIP diff for removing `create-react-app` from the recorder and replacing it with dependencies that we will fully maintain. We'll be pulling changes into this branch from our forks until we get to a point where the code is in line with our goals.

## Implementation details

We ran the [eject](https://github.com/facebook/create-react-app/blob/main/packages/react-scripts/scripts/eject.js) script to get a baseline working solution. Goals:

1. Get this branch passing CI
1. Make incremental revisions to this branch via PRs from our forks
1. When we are feeling good about the config maintenance burden, we will review and merge the overall patch

Ensure the build/package scripts are working and that you can run the recorder.


justinkambic:
There's been no activity on this for some time, closing for now. If we decide to re-visit in the future we can re-open.

PR #231, opened 2022-04-28
chore: Move `createAction` test helpers to top level of repo

### LINKED ISSUE #230: [Other] Move test helpers to top of repo
### Task summary

We have some test helpers, `createAction`, `createSteps`, etc., that can be employed in the electron code as well. Thus, we can move them to the top of the repo.

### Linked resources

_No response_

### Additional information

_No response_

justinkambic:
## Summary

Resolves #230.

Moves some helper functions.

## Implementation details

I wrote this code as a part of #211, but as they're not really related to that fix I have instead moved the commit to a separate PR.

All unit tests should pass.


kyungeunni (commented):
Great work Justin 👍  I noticed a couple of things.

1. The error message with the code preview is not visible on the flyout when test failed:
image

2. It is not reliably reproducible, but I've encountered this case when deleting and adding some steps, the test would fail and this error was printed in the console. I will try to find a way to reproduce it.
image

kyungeunni (approved):

PR #233, opened 2022-04-28
fix: Error not showing in failed test result flyout

justinkambic:
## Summary

Fixes a defect we introduced recently as part of the backend transition from pure JS to TypeScript. I'm going to add tests for this function so we don't fail to notice things like this in the future.

### Before

image

### After

image

## Implementation details

We simply forgot to include an optional field in the response from the test procedure.

Run a journey that you expect to fail, and observe the code/data info in the test result flyout.


kyungeunni (approved):
Oops. My bad. thanks for fixing it so quickly 👍

justinkambic:
Issue to add tests: https://github.com/elastic/synthetics-recorder/issues/235.

PR #241, opened 2022-05-04
feat: Add a controller class for browser process requests

### LINKED ISSUE #149: [Enhancement] Better handling on recording or running journey
### Feature summary

It shouldn't be possible for users to run and record the journey at the same time. As mentioned in this [comment](https://github.com/elastic/synthetics-recorder/pull/143#discussion_r800895161) it is possible to record a new journey while the test is still running. It is fixed within the UI by https://github.com/elastic/synthetics-recorder/commit/80fe8e3036e6e687e644487cfd604f7e22f4bef3 but I'd like to suggest we also handle it in API level too.

Currently, all the API calls from the frontend will be executed simultaneously as they are emitted from individual listeners and they don't know about others' states. So it is possible to run and record the journey at the same time and it is not desired behaviour and should be prevented at all.

 As an extreme example, it is possible that when several [`run-journey`](https://github.com/elastic/synthetics-recorder/blob/main/electron/execution.js#L270) events are emitted at the same time, it can [spawn](https://github.com/elastic/synthetics-recorder/blob/main/electron/execution.js#L195) several synthetics cli processes as there is no throttling and control over it(thankfully it is prevented in the UI but it _can_ happen).

It would be beneficial to introduce some sort of protection around it, possibly introducing _Journey_ domain controller(or similar) to keep track of the requests from the frontend and throttle if necessary.

### Linked resources

relates to https://github.com/elastic/synthetics-recorder/issues/127

### Additional information

_No response_

justinkambic:
## Summary

Resolves #149.

## Implementation details

Adds a class for managing browser process handlers. We have already disallowed the client to send test/record requests while there's already a session in progress. The backend code does not regulate this, however.

This patch adds a new class that will keep a reference to a current request, and clear the request after it resolves. Any subsequent requests will log a warning and call an IPC failure channel. Errors in the process are likewise logged before the process buffer is cleared.

At the moment the class only buffers a single process. It will be easy to extend to holding additional processes in the future if we want to add functionality that requires it.

Ensure that you cannot start multiple browser-dependent processes at the same time.

You can remove the check if a test is in progress from the Start Button's [`disabled`](https://github.com/elastic/synthetics-recorder/blob/main/src/components/Header/HeaderControls.tsx#L90) prop.

1. Record a long journey
2. Run a test
3. Attempt to start recording
4. You should be able to see a warning logged that there's already a process running

### DIFF
--- a/electron/BrowserRequestController.ts
+++ b/electron/BrowserRequestController.ts
@@ -1,0 +1,81 @@
+/*
+MIT License
+
+Copyright (c) 2021-present, Elastic NV
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
+import { BrowserWindow } from 'electron';
+import { ipcMain as ipc } from 'electron-better-ipc';
+import logger from 'electron-log';
+import type { ClientBrowserRequest, RecordJourneyRequest } from '../common/types';
+import { isTestJourneyRequest } from '../src/common/shared';
+
+const isRecordJourneyRequest = (req: any): req is RecordJourneyRequest => !!req?.data?.url;
+
+export class BrowserRequestController {
+  private buffer: ClientBrowserRequest | null;
+
+  constructor() {
+    this.buffer = null;
+  }
+
+  public async executeRequest<T>(
+    req: ClientBrowserRequest,
+    browserWindow: BrowserWindow,
+    execution: (req: ClientBrowserRequest, browserWindow: BrowserWindow) => Promise<T>
+  ): Promise<T | undefined> {
+    const isRecordingRequest = isRecordJourneyRequest(req);
+    const isTestRequest = isTestJourneyRequest(req);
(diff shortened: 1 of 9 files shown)


electron/BrowserRequestController.ts
```diff
@@ -0,0 +1,81 @@
...
+          'Cannot start recording a journey, a browser operation is already in progress.'
+        );
+      } else if (isTestRequest) {
+        logger.warn('Cannot start testing a journey, a browser operation is already in progress.');
+      }
+      // drop any requests we receive while a browser is running
+      return ipc.callRenderer(browserWindow, 'browser-request-failure', req);
+    }
+
+    this.buffer = req;
```
kyungeunni, line 59: I don't see this saved payload used somewhere else - is there a reason you chose to save data instead over boolen?

electron/BrowserRequestController.ts
```diff
@@ -0,0 +1,81 @@
...
+
+import { BrowserWindow } from 'electron';
+import { ipcMain as ipc } from 'electron-better-ipc';
+import logger from 'electron-log';
+import type { ClientBrowserRequest, RecordJourneyRequest } from '../common/types';
+import { isTestJourneyRequest } from '../src/common/shared';
+
+const isRecordJourneyRequest = (req: any): req is RecordJourneyRequest => !!req?.data?.url;
+
+export class BrowserRequestController {
```
kyungeunni, line 33: It seems like what this class does is just to keep track of whether the browser is running or not. Could this whole class be replaced by a boolean variable in execution?
```js
// electron/execution.ts
let isBrowserRunning= false;

async function onRecordJourneys(...args) {
  if (isBrowserRunning) {
   throw new Error('Cannot start recording a journey, a browser operation is already in progress.');
  } 
  isBrowserOpen = true;
   ...
}

async function onTest(...args) {
  if (isBrowserRunning) {
   throw new Error('Cannot start testing a journey, a browser operation is already in progress.');
  } 
  isBrowserOpen = true;
   ...
}
```

But please let me know if there are some important points that I missed 👍

vigneshshanmugam: ++ felt like we are over complicating things here by using controller class.

justinkambic: I may have misinterpreted the original intention behind the issue. Certainly we can achieve this with a simple boolean instead, sorry for over-engineering this.

electron/BrowserRequestController.ts
justinkambic: I think I misinterpreted the intention of the original issue; I was designing this in a way that we'd be able to queue additional jobs and track their state in the future if we wanted to support more than one simultaneous child process.

justinkambic:
@kyungeunni I have updated the change based on your feedback, LMK what you think. It is much simpler now.

electron/BrowserRequestController.ts
kyungeunni: Sorry for the misleading issue - I should've revisited it and updated it as some of the assumptions/suggestions I had back then were not that intuitive and well-thought through.

kyungeunni (approved):
LGTM 👍 
Just wanted to say sorry again for the misleading issue! 🙏 
(I did smoke tests and confirmed it wouldn't open more than one browser)

electron/BrowserRequestController.ts
justinkambic: No worries. I have pushed the commits with the class/tests to my remote so if we ever want to enhance this limiting to allow simultaneous test runs/recording sessions we can draw on that code for inspiration.

PR #245, opened 2022-05-10
fix: Kill child processes when main window closes

### LINKED ISSUE #243: [Bug] Child recorder process is not terminated when recorder window closes
### Bug summary

When the user closes the recorder's main window, it allows its child browser process to continue to run. After the child window is closed, an error message is displayed. In dev mode, the recorder simply logs the message to console:

```
[2] 12:56:00.220 › TypeError: Object has been destroyed
[2]     at ~/synthetics-recorder/node_modules/electron-better-ipc/source/main.js:46:20
[2]     at new Promise (<anonymous>)
[2]     at IpcMainImpl.ipc.callRenderer (~/synthetics-recorder/node_modules/electron-better-ipc/source/main.js:9:54)
[2]     at EventEmitter.actionsHandler (~/synthetics-recorder/build/electron/execution.js:92:39)
[2]     at EventEmitter.emit (events.js:376:20)
[2]     at CodeGenerator.didPerformAction (~/synthetics-recorder/node_modules/playwright/lib/server/supplements/recorder/codeGenerator.js:126:138)
[2]     at CodeGenerator.addAction (~/synthetics-recorder/node_modules/playwright/lib/server/supplements/recorder/codeGenerator.js:69:10)
[2]     at Page.<anonymous> (~/synthetics-recorder/node_modules/playwright/lib/server/supplements/recorderSupplement.js:303:23)
[2]     at Page.emit (events.js:388:22)
[2]     at Page._didClose (~/synthetics-recorder/node_modules/playwright/lib/server/page.js:155:10)
```

In production builds, the message shows as much more prominent modal window:

image

### Recorder Version

alpha-4

### Steps to reproduce

1. Run the recorder
2. Start a recording session
3. Kill the recorder window using the close button or CMD/CTRL+W
4. Kill the child browser window
5. Observe the error

### Expected behavior

There should be no error message because the parent window should kill the child process as part of its `close` routine.

### Additional information

_No response_

justinkambic:
## Summary

Resolves #243.

## Implementation details

I've added an event emitter to the main electron file. We provide this emitter to our listeners when they're bootstrapped, so any handler that needs to listen to the main thread is now a function factory that provides this emitter to the handler function.

The handler function then listens to the emitter for a `close` event, and proceeds to do the necessary cleanup, killing browser context or child processes to ensure no references are left dangling and causing poor user experiences.

There are two main issues this patch fixes.

1. When the user closes the main window while recording, the chromium child process does not terminate.
2. When the user closes the main window while recording or testing, the underlying JS code can attempt to reference or utilize objects that are destroyed or in an invalid state.

### Recorder issue

While recording, if the user closes the main window, the chromium should close and there should be no errors logged or warning windows displayed (see the linked issue for more info).

### Testing issue

While running a test, if the user closes the main window, the test chromium should also close as it serves no purpose without the recorder being open

### Additional context

Neither of these changes should have an impact on the functioning of the application when subsequent windows are opened/closed. Subsequent windows of the recorder should be able to fully function normally. It is possible that testing this is blocked by https://github.com/elastic/synthetics-recorder/pull/192.

### DIFF
--- a/electron/execution.ts
+++ b/electron/execution.ts
@@ -333,7 +333,9 @@ function onTest(mainWindowEmitter: EventEmitter) {
       });
     } finally {
       if (synthCliProcess && !synthCliProcess.kill()) {
-        logger.warn("Couldn't kill browser process.");
+        logger.warn(
+          `Attempted to send SIGTERM to synthetics process, but did not receive exit signal. Process ID is ${synthCliProcess.pid}.`
+        );
       }
     }
   };


electron/electron.ts
```diff
@@ -41,7 +42,10 @@ const BUILD_DIR = join(__dirname, '..', '..', 'build');
 const IS_TEST = process.env.NODE_ENV === 'test';
 const TEST_PORT = process.env.TEST_PORT;
 
+let mainWindowEmitter: EventEmitter | null = null;
```
lucasfcosta, line 45: Instead of making this emitter global, can we return it from `createWindow` and  use it in the `then` closure for `appWhenReady` below?

electron/electron.ts
```diff
@@ -71,7 +82,10 @@ function createMenu() {
 
 app.whenReady().then(() => {
   createWindow();
```
lucasfcosta, line 84: Here we could return the emitter and therefore we wouldn't have to do the `null` check, reassign after removing listeners, or even have the global reference.

lucasfcosta (changes requested):
Hey Justin, thanks for this PR!

I had minor recommendations wrt to the reference to `mainWindowEmitter` but the main reason I'm requesting changes is that I've found the following bug on MacOS:

* When closing the main window during testing, the browser won't open again if you reopen the main window and try to record a test.

That could be confusing because users could wonder why they can't record anything after having closed the main window.

It's a very "specific" bug in a way, so please let me know if you think it's not worth tackling now.

Here's how to repro:

1. Open the recorder and record a journey
2. Close the browser window
3. Click "test"
4. While the test is running, close the main recorder window
5. Reopen the main window through the bottom tray
6. Try to record a new test and you'll see the browser window won't open.

I think that's because you only properly quit the recorder with CMD+Q, not by closing the main window, so we may need to reset state there too.

electron/electron.ts
justinkambic: Great recommendation, thank you!

electron/electron.ts
justinkambic: Perfect 🙌

justinkambic:
Great bug catch Lucas, thanks for this review. I am pretty sure I know exactly what's wrong here, and it's actually happening on `main` as well. I am going to push another PR to fix that problem separately, then will rebase/update here when the fix is merged.

vigneshshanmugam:
Nothing against the current approach, There is one other way to do this which I have done in the past without using EventEmitter pattern. 

```js
const childProcess  = []

childProcess.push(browser, synthetics);

app.on("exit", childProcess.forEach(process => process.kill());

```

justinkambic:
I guess it's more of an architecture decision - to keep a list of child processes as shown we'll be centralizing them at the top level. The main drawback to this approach in the context of solving the original issue is that there's other cleanup we need to do besides just killing the processes to avoid error output (i.e. removing listeners before killing a browser process), so I felt it better to propagate the termination events down to the handler code. If we used that approach in the `execution.ts` file we'd still need to propagate the close event downward.

lucasfcosta (changes requested):
Hey Justin, thanks for the update!

I've just pulled this branch and now I'm facing a different problem: it seems that the browser will _not_ be closed when I close the main window during the `Test` nor when I close it during the recording process. 😢

lucasfcosta (approved):
LGTM as per the comment above. Also tested reopening the main window and trying to re-record once I had closed it during a test.

electron/execution.ts
```diff
@@ -90,51 +94,58 @@ let browserContext: BrowserContext | null = null;
...
+      browserContext = context;
+      actionListener = new EventEmitter();
+      actionListener.on('actions', actionsHandler);
 
-    ipc.on('stop', closeBrowser);
+      mainWindowEmitter.addListener(MainWindowEvent.MAIN_CLOSE, async () => {
+        actionListener.removeAllListeners();
+        ipc.removeListener('stop', closeBrowser);
+        return browser.close().then(() => {
+          isBrowserRunning = false;
```
kyungeunni, line 129: Not so important comment - as the function is async, shall we use await here?
```suggestion
       await browser.close();
       isBrowserRunning = false;
```

electron/execution.ts
```diff
@@ -90,51 +94,58 @@ let browserContext: BrowserContext | null = null;
...
+      // Listen to actions from Playwright recording session
+      const actionsHandler = (actions: ActionInContext[]) => {
+        ipc.callRenderer(browserWindow, 'change', { actions });
+      };
+      browserContext = context;
+      actionListener = new EventEmitter();
+      actionListener.on('actions', actionsHandler);
 
-    ipc.on('stop', closeBrowser);
+      mainWindowEmitter.addListener(MainWindowEvent.MAIN_CLOSE, async () => {
```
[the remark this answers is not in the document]
kyungeunni: I can see it will be removed when the window is closed, but I'm a bit worried we will be keep adding the listeners whenever we record and test a script

electron/electron.ts
```diff
@@ -62,25 +64,41 @@ async function createWindow() {
...
 function createMenu() {
   const menuTemplate = buildMenu(app.name);
   Menu.setApplicationMenu(Menu.buildFromTemplate(menuTemplate));
 }
 
-app.whenReady().then(() => {
-  createWindow();
-  setupListeners();
-  createMenu();
+async function mainWindow() {
```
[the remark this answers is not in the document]
justinkambic: fee2aa07a84d5868e9f5a5596563a9bb1a87ec3b

electron/execution.ts
justinkambic: 0278c46d9e0c345560c521c8707907e6d62c94d4

kyungeunni (approved):
I thought I pressed approve button :D

electron/execution.ts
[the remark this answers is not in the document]
justinkambic: 9e176847e7f7aa8031fd4fa1272b5cc59ec1828b

electron/execution.ts
```diff
@@ -165,147 +176,159 @@ function addActionsToStepResult(steps: Steps, event: StepEndEvent): TestEvent {
...
+          PLAYWRIGHT_BROWSERS_PATH,
         },
-      };
-    }
-    if (isJourneyEnd(parsed)) {
-      return {
+        cwd: isDev ? process.cwd() : process.resourcesPath,
+        stdio: 'pipe',
+      });
+      mainWindowEmitter.addListener(MainWindowEvent.MAIN_CLOSE, () => {
```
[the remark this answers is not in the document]
justinkambic: 9e176847e7f7aa8031fd4fa1272b5cc59ec1828b

justinkambic:
@kyungeunni thanks for those comments - made the code even more prod-worthy ❤️

PR #247, opened 2022-05-12
feat: Add soft delete to steps functionality

### LINKED ISSUE #242: [Enhancement] Be able to delete actions during the recording
### Feature summary

As a user of the Script Recorder
I want to be able to delete an action whilst recording
So that I can remove unnecessary browser interactions

ACs:
* I can delete an action even if the recording is in progress, or paused

When using the script recorder, one can easily click on things by mistake (e.g. identifying a selector), which creates unnecessary _noise_ in the recording.

It’s not currently possible to delete these whilst recording, and although it is possible once the recording is stopped, it can be be difficult to remember which were the unnecessarily captured actions, once finished (i.e. it’s easier to be aware of them when the additional actions are captured rather than at the end).

### Linked resources

_No response_

### Additional information

_No response_

justinkambic:
## Summary

Resolves #242.

## Implementation details

Adds an `isSoftDeleted` flag to the `ActionContext` type that we use to represent custom IR values. When a user clicks the `Delete` button while recording, the recorder will apply this flag to the targeted action. ~After recording ends, the recorder will filter that action out of the list.~ Additionally, we will set the `modified` flag on any deleted action. This will cause the existing [`generator` code](https://github.com/elastic/synthetics-recorder/pull/247/files#diff-7cc5d95f5c01783f59eac8f7333a807ce5a224ce3d94d0899aba6d531e120b80R100) to preserve the deleted state when Playwright sends new actions. We need to keep these `modified` actions because even after the Recorder's internal state reflects that recording has stopped, it can still handle latent Playwright events that would overwrite the state; because of this, we can't reliably modify the state to remove the items where `isSoftDeleted == true`.

The export flyout also checks for this flag when generating code. ~This is worth some additional examination, it might be better to modify the code that actually calls the main thread to perform this check on its own, so that we can remove this added abstraction from the UI component layer where possible.~

1. Start recording a journey
2. "Delete" an action
3. The action element should disappear
4. Open the export flyout, you should see no code associated with the action you deleted
5. Stop recording and log the steps in `App.tsx`, you should not see any of the actions you soft-deleted while recording


src/common/shared.ts
```diff
@@ -92,26 +95,6 @@ export function createExternalLinkHandler(
   };
 }
 
-export function updateAction(
```
justinkambic, line 95: Deleted because this function is not used in the codebase anywhere.

src/components/ActionElement/ExtraActions.tsx
```diff
@@ -78,6 +79,7 @@ export function ExtraActions({
       </EuiFlexItem>
       <EuiFlexItem grow={false}>
         <SettingsPopover
+          isAssertion={actionContext.action.isAssert === true}
```
lucasfcosta, line 82: Nit:

```suggestion
          isAssertion={actionContext.action.isAssert}
```

src/hooks/useRecordingContext.ts
```diff
@@ -52,12 +53,19 @@ export function useRecordingContext(
       setRecordingStatus(RecordingStatus.NotRecording);
       // Stop browser process
       ipc.send('stop');
+      // remove any actions that were soft-deleted while recording
+      setSteps(steps =>
+        steps.map(step => ({
+          ...step,
+          actions: step.actions.filter(action => (action as ActionContext)?.isSoftDeleted !== true),
```
lucasfcosta, line 60: Nit:

```suggestion
          actions: step.actions.filter(action => !(action as ActionContext)?.isSoftDeleted),
```

lucasfcosta (changes requested):
Thanks for this one, @justinkambic!

Just tested this and the functionality still seems not to be working 🤔 

See in the gif below that when the browser is closed the deleted steps are restored.

no_delete

Also, can we make sure to add comments to the code itself as to why we need to "soft" delete? You and I have talked about that in our call last week, and there's also info in this PR, but I'm afraid it may get lost in the context of further changes.

src/components/ActionElement/ExtraActions.tsx
justinkambic: This will break TS as `isAssert` is optional, but `isAssertion` requires a `boolean`. We could modify `isAssertion` to accept `undefined`, but that pushes the complexity into the implementation of `SettingsPopover`.

Alternatively, we could write something like `!!actionContext.action.isAssert`. I felt like the explicit check against `true` was fine, but I'm open to whatever you'd find most readable.

lucasfcosta: Ah, that makes sense! Thanks for taking the time to explain. We can keep as is.

src/helpers/generator.test.ts
```diff
@@ -527,5 +527,134 @@ describe('generator', () => {
...
+                modifiers: 0,
+              },
+              committed: true,
+              title: 'Press Enter',
+            },
+            {
+              pageAlias: 'page',
+              isMainFrame: true,
+              frameUrl:
+                'https://www.google.com/search?q=hello+world&source=hp&ei=Wa-wYt7xDPa15NoP6NeGmAk&iflsig=AJiK0e8AAAAAYrC9aQ-K5PLjczC0c7dceKI--pjoitdO&ved=0ahUKEwieguz8x7z4AhX2GlkFHeirAZMQ4dUDCAo&uact=5&oq=hello+world&gs_lcp=Cgdnd3Mtd2l6EAMyCAgAEIAEELEDMgsILhCABBCxAxDUAjIICAAQgAQQsQMyCAgAEIAEELEDMggIABCABBCxAzIFCC4QgAQyCAgAEIAEELEDMggIABCABBCxAzILCC4QgAQQsQMQ1AIyBQgAEIAEOg4IABCPARDqAhCMAxDlAjoOCC4QjwEQ6gIQjAMQ5QI6EQguEI8BENQCEOoCEIwDEOUCOhEILhCABBCxAxCDARDHARCjAjoLCC4QgAQQxwEQrwE6CAgAELEDEIMBOgQIABADOgsIABCABBCxAxCDAToICC4QsQMQgwE6DgguEIAEELEDEMcBEKMCOhEILhCABBCxAxCDARDHARCvAToOCC4QgAQQxwEQrwEQ1AI6DgguELEDEIMBEMcBEKMCOhEILhCABBCxAxCDARDHARDRAzoLCC4QgAQQsQMQgwE6CAguEIAEELEDOg4IABCABBCxAxCDARDJAzoQCC4QgAQQsQMQxwEQ0QMQCjoICAAQgAQQyQM6BwgAELEDEApQrwJYvAtgsgxoAXAAeACAAVeIAbgFkgECMTGYAQCgAQGwAQo&sclient=gws-wiz',
```
lucasfcosta, line 617: can we remove these long URLs and just use dummy ones?

lucasfcosta (approved):
LGTM, also manually tested this PR by:

1. Starting a recording.
2. Visiting a website and generating new actions by clicking around
3. Deleting one of them during the recording.
4. Closing the browser and verifying actions and exported code.

deleting_rec

I do think we should really remove those long URLs from the tests though. Perhaps we could create a function to generate those to make data there more succint?

src/helpers/generator.test.ts
justinkambic: yes, that's no problem. These URLs have no material impact and it will be much cleaner. In fact, we have the test object generators, I should have used them in the first place. I'll push a commit to accomplish this.

PR #249, opened 2022-05-17
feat: Add option to build `arm64` binary

### LINKED ISSUE #248: [Enhancement] Ship `arm64` and `x64` binaries for macOS
### Feature summary

As part of #52, we wanted to ship a universal build for the recorder. There are a few reasons that preclude us from achieving this, so for the time being we will ship the recorder with separate Intel/Apple silicon binaries.

As a side benefit, these separate binaries will be significantly smaller than the universal build by a factor of close to 100%.

The recorder runs faster and has a snappier UI experience when build for native arm systems, so by providing this option out of the box users of newer Apple machines will have a nicer time using the recorder.

### Linked resources

_No response_

### Additional information

_No response_

justinkambic:
## Summary

Resolves #248.

Causes the `build`/`release` scripts to generate an Apple silicon binary in addition to the Intel one.

## Implementation details

Simply updates the build config for the recorder.

You can do an `npm run pack` command to build fresh binaries locally. You should be able to run both the `x64` and the `arm64` versions without issues.

### DIFF
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -19,6 +19,9 @@ mac:
   extraResources:
     - from: local-browsers/_releases/mac
       to: local-browsers
+  target:
+    target: default
+    arch: [x64, arm64]
 win:
   target: nsis
   icon: public/elastic.png
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -1,5 +1,6 @@
 productName: "Elastic Synthetics Recorder"
 appId: co.elastic.synthetics-recorder
+artifactName: ${productName}-${version}-${os}-${arch}.${ext}
 beforePack: "./scripts/before-pack.js"
 afterPack: "./scripts/after-pack.js"
 files:
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -5,6 +5,9 @@ afterPack: "./scripts/after-pack.js"
 files:
   - build/**/*
   - scripts/**/*
+extraResources:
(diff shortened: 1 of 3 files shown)


kyungeunni:
@justinkambic After our call, I found a way to include the correct browser directory per build. It should download and bundle the correct browser per os/architecture including Apple Silicon.

electron-builder.yml
```diff
@@ -5,6 +5,9 @@ afterPack: "./scripts/after-pack.js"
 files:
   - build/**/*
   - scripts/**/*
+extraResources:
```
justinkambic, line 8: Wow great work this is much cleaner!

scripts/download-chromium.js
```diff
@@ -80,15 +85,15 @@ function translatePlatform(platform) {
   }
 }
 
-exports.downloadForPlatform = async function downloadForPlatform(platform) {
-  platform = translatePlatform(platform);
+exports.downloadForPlatform = async function downloadForPlatform(electron_platform, arch) {
```
justinkambic, line 88: Nice name update.

kyungeunni:
Addressed the suggested artifact naming improvements!! Now all the artifacts have the same name format including the arch. 
image

kyungeunni:
Hey @justinkambic I think this is ready to be reviewed, then merged if you are happy with it!

kyungeunni (approved):

PR #253, opened 2022-05-19
Update all NPM deps

andrewvc:
We've angered dependabot, this should make it happier and bring things closer to current.

Seems to work with `npm run dev`, will wait for green tests here as well.


justinkambic (commented):
I've tested this locally and it seems to build/run ok and all tests pass. When I did an `npm run pack` however the outputted binary didn't seem to have the right chromium for playwright, when I tried recording nothing happened.

I have been doing some fiddling with my local setup for another feature, so it's possible that my cache is dirty or some other issue is making the pack not work. @kyungeunni if you can try doing a pack from this branch as well and LMK if you see that issue, other wise I'm fine to approve.

kyungeunni:
LGTM 👍 @justinkambic I've tested both `npm run dev` and binaries generated by `npm run pack` from this branch, and all worked fine for me. I noticed it can be _very slow_ to see the browser up and running when it's the first time you launch the browser from the app, so it could be it was just being slow?

kyungeunni (approved):

justinkambic:
> I noticed it can be very slow to see the browser up and running when it's the first time you launch the browser from the app, so it could be it was just being slow?

It's possible but the UI was reflecting that the session ended. The Play/Pause button reverted to a Play icon, which led me to conclude that the browser crashed/failed to start/was not there.

I have also noticed the browser is extremely slow to start on its initial launch. I'm unsure if we can do anything about improving the launch speed.

vigneshshanmugam (approved):
Tested on M1, looks good 👍🏽

shahzad31:
Dependencies are updated, skipping post FF Testing

PR #256, opened 2022-05-20
fix: Browser will not launch after manual termination of test run

### LINKED ISSUE #255: [Bug] Manually closed before never gets respawned
### Bug summary

+ During the middle of the test run, If the browser is closed manually, Users wont be able to reset and record new tests

### Recorder Version

main

### Steps to reproduce

1. Start the Recorder
2. Type in URL and start and finish recording
3. Press `Test` button and close the browser manually
4. Now press `Start` and try to delete old recorder steps and Start over
5. Browser window will not be opened and UI will be in the `Recording` state which is incorrect. 

### Expected behavior

_No response_

### Additional information

_No response_

justinkambic:
## Summary

Resolves #255.

Check the issue above for detailed repro steps - essentially if the user terminates the PW browser during a test run, the Recorder can't open the browser back up again ever.

## Implementation details

We [recently introduced ](https://github.com/elastic/synthetics-recorder/pull/241/files#diff-33107fe7222398333570c06513ba085b86bd3564c21860f8a70886736dff81e6R91) an `isBrowserRunning` flag to prevent the recorder from launching unbounded browser instances.

Unfortunately, I failed to ensure that the flag would be reset in all cases. This change should correct that. It nests all setting of the `isBrowserRunning` flag inside of the pre-existing `try...catch...finally` blocks that the affected functions contain; whether the browser terminates manually or as planned, these flags should be reset afterwards.

Use the repro steps to confirm the problem on `main`, and then checkout and run this branch to observe that this patch resolves the problem.

### DIFF
--- a/electron/execution.ts
+++ b/electron/execution.ts
@@ -144,8 +144,9 @@ function onRecordJourneys(mainWindowEmitter: EventEmitter) {
       mainWindowEmitter.removeListener(MainWindowEvent.MAIN_CLOSE, handleMainClose);
     } catch (e) {
       logger.error(e);
+    } finally {
+      isBrowserRunning = false;
     }
-    isBrowserRunning = false;
   };
 }
 
@@ -335,8 +336,8 @@ function onTest(mainWindowEmitter: EventEmitter) {
           `Attempted to send SIGTERM to synthetics process, but did not receive exit signal. Process ID is ${synthCliProcess.pid}.`
         );
       }
+      isBrowserRunning = false;
     }
-    isBrowserRunning = false;
   };
 }
 
--- a/electron/execution.ts
+++ b/electron/execution.ts
@@ -101,8 +101,8 @@ function onRecordJourneys(mainWindowEmitter: EventEmitter) {
         'Cannot start recording a journey, a browser operation is already in progress.'
       );
     }
+    isBrowserRunning = true;
     try {
-      isBrowserRunning = true;
       const { browser, context } = await launchContext();
       const closeBrowser = async () => {
         browserContext = null;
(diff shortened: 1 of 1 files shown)


kyungeunni (approved):
It fixed mentioned bug, thanks!

p.s. It was quite difficult to read the function and locate where the changes are applied or what part of the function has changed. It seems like it's because the function is too large and it contains so many nested functions inside. It'd be great if we can tidy up a little bit when we have time. I'll create a ticket

electron/execution.ts
```diff
@@ -283,6 +281,9 @@ function onTest(mainWindowEmitter: EventEmitter) {
         await writeFile(filePath, data.code);
         args.unshift(filePath);
       }
+
+      isBrowserRunning = true;
```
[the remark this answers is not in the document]
justinkambic: I can move it back where it was previously.

electron/execution.ts
```diff
@@ -125,8 +125,7 @@ function onRecordJourneys(mainWindowEmitter: EventEmitter) {
       const handleMainClose = async () => {
         actionListener.removeAllListeners();
         ipc.removeListener('stop', closeBrowser);
-        await browser.close();
-        isBrowserRunning = false;
+        return browser.close();
```
vigneshshanmugam, line 128: can you throw errors in this listener and check if the `isBrowserRunning` variable is rest in the finally block?

vigneshshanmugam (approved):
Tested, can startover the recording. LGTM, but would like to validate one change which is mentioned on the review.

electron/execution.ts
justinkambic: If I understand this code correctly any errors from this handler should get thrown on their own right? `isBrowserRunning` is reset in the finally block by default already.

PR #259, opened 2022-05-24
ci: upload exe files

v1v:
## Summary

Upload exe files to the google bucket so the release will consume those binaries


kyungeunni (approved):
LGTM!!

PR #260, opened 2022-05-30
fix protocol build config

shahzad31:
fix protocol build config, protocols key takes an object instead of array 

originally this was defined in package.json as json, but when it got moved to yaml format in this PR https://github.com/elastic/synthetics-recorder/pull/165

protocols got defined as array, which means it stopped working.

### DIFF
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -14,9 +14,9 @@ publish:
 extraMetadata:
   main: build/electron/electron.js
 protocols:
-  - name: elastic-synthetics-recorder
-    schemes:
-      - elastic-synthetics-recorder
+  name: elastic-synthetics-recorder
+  schemes:
+    - elastic-synthetics-recorder
 mac:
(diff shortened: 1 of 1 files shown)


kyungeunni:
Thanks, @shahzad31! 
According to electron-builder doc, array also should be accepted:
image

Also, it seems to work for me(the build from the current main branch), could you double-check, please?
[Image from Gyazo](https://gyazo.com/c8af3cbe9a56fc0f652059500bdce830)

shahzad31:
@kyungeunni in that case we can disregard this PR :)

i was actually testing adding a link in kibana to open script recorder and it wasn't working. not sure what's going on there.

kyungeunni:
I noticed a weird behaviour actually, it works only when the app is fully closed

shahzad31:
> I noticed a weird behaviour actually, it works only when the app is fully closed

interesting, might be worth investigating

PR #238, opened 2022-05-03
ci: move release process to the internal system

v1v:
## Summary

Enable tag release to run in the internal system, so the artifacts are signed and published in GitHub.

## Implementation details

All the previous stages will do nothing if running on tag releases in the `apm-ci`.

The signing happens as part of the internal toolchain, therefore it's required to generate the packages and sign them

## How to validate this change

In the CI

## Questions

- Is it possible to generate the packages on Linux?

## Tasks

- [ ] Sign macos 
- [ ] Sign windows
- [ ] Sign linux artifacts

PR #261, opened 2022-05-31
fix: remove dashes in app name

kyungeunni:
## Summary
The latest builds have an app name "Elastic-Synthetics-Recorder". Before this, it was "Elastic Synthetics Recorder" (the current build has dashes between the words.)

image

## Implementation details
In `electron-build.yml`, revert the `productName` field to have empty spaces instead of dashes. At the same time, in order to prevent any issue with empty space in artifacts name, we will use the `name` field from `package.json` instead of productName which does not contain empty space at all times. 

I'd love to check with @v1v about the change introduced [here](https://github.com/elastic/synthetics-recorder/pull/238/files#diff-abfd13c4441842dc2fc303311c40908ecd004ecec4c4bb5827c366b454a1f721L1) was due to the empty spaces in the productName having side effect when building artifacts. (If that was the only reason I've configured the electron-builder to use another field that doesn't have any empty space for artifacts name).

1. run `npm run pack -- -mwl` locally.
2.Go to `dist` directory and check the unpacked apps do not contain dashes in their name:
image
or install the app and confirm:
image

3. Also check all the builds have the same naming convention.(if artifacts name had an empty space, it would have different format as it was in the previous [releases](https://github.com/elastic/synthetics-recorder/releases/tag/v0.0.1-alpha.3): 
image

### DIFF
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -1,6 +1,6 @@
-productName: "Elastic-Synthetics-Recorder"
+productName: "Elastic Synthetics Recorder"
 appId: co.elastic.synthetics-recorder
-artifactName: ${productName}-${version}-${os}-${arch}.${ext}
+artifactName: ${name}-${version}-${os}-${arch}.${ext}
 beforePack: "./scripts/before-pack.js"
 afterPack: "./scripts/after-pack.js"
 files:


v1v (approved):
LGTM, if the artifact name does not contain any spaces.

The original change was done to actually remove spaces in the artifact name, somehow I missed the `productName` was also used somewhere else 🙇

justinkambic (approved):
LGTM!

PR #273, opened 2022-06-21
fix: Update electron-builder to get later version of `caniuse`

justinkambic:
## Summary

I've noticed that ci seems to be failing with error messages related to `caniuse` being outdated. I'm bumping electron builder here to see if this resolves the problem.


justinkambic:
Superseded by #275.

PR #275, opened 2022-06-24
fix: update deps to fix builds

kyungeunni:
One of the sub-dependencies is causing the build to fail. This PR tries to fix by installing a later version of the problematic dependency, but ultimately we would love to have better control over dependencies by addressing issue #220 

## What does this PR include
1. Update some of the dependencies by running `npm run update`
2. Set later version of `caniuse-light` in package.json > resolutions which satisfy `browserslist`(used in `react-script > postcss-preset-env > browserslist`)
3. Update jest and ts-jest to latest version as the currently used version throws UnhandledPromiseRejectionWarning so when the build fails
4. Tweak jest configs to address some of the breaking changes introduced by 27.x and 28.x

Before(jest@26.x):
Screenshot 2022-06-24 at 19 28 34
After (jest@28.1.1):
Screenshot 2022-06-24 at 19 29 53

## How to verify the fix
The build should be successful and then all the CI should pass

## Nest steps
I'd like to rearrange tests-related resources under `tests` dir that is introduced in this PR, namely `e2e/*` and `demo-app/*` should be moved to `tests` dir so it is clear that they are not related to the application's main functionality.


tests/common/resolver.js
```diff
@@ -0,0 +1,55 @@
...
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
+// Original work: https://github.com/microsoft/accessibility-insights-web/blob/main/src/tests/common/resolver.js
```
justinkambic, line 25:
```suggestion
// Original work: https://github.com/microsoft/accessibility-insights-web/blob/871cd3ab84833a79bcc4c53d58f51b8bac23200a/src/tests/common/resolver.js
```

Recommend we include a commit hash so this link doesn't die if the file is ever moved or deleted.

justinkambic (approved):
Tested this locally and changes LGTM. Thanks for unblocking our CI ❤️

PR #280, opened 2022-07-04
chore: rename `suites` to `projects`

### LINKED ISSUE #254: Rename `suites` to `projects`
AC:
* Change `suites` to `projects` (that's the scope of this issue)

image

Original issue text below

~We currently let users export javascript for both inline scripts and projects via the UI shown below. The ability to format for projects is confusing to some users. This issue kicks off a discussion on how we can improve this.~

~The more I think about it, the more I think we don't need that feature, for script users they can add their own wrapping `journey statements`, but I want to leave this issue up as a placeholder.~

kyungeunni:
closes #254 

## Summary
Aligning the usage of terminology within Synthetics product. We decided to use `Projects` instead of `Suites`.

Updated all the Suite usage both in code and visual contents so that we are not confused by deprecated terms in future development.

### DIFF
--- a/src/components/ExportScriptFlyout/Body.tsx
+++ b/src/components/ExportScriptFlyout/Body.tsx
@@ -28,18 +28,18 @@ import type { Setter } from '../../common/types';
 
 interface Props {
   code: string;
-  exportAsSuite: boolean;
-  setExportAsSuite: Setter<boolean>;
+  exportAsProject: boolean;
+  setExportAsProject: Setter<boolean>;
 }
 
-export function Body({ code, exportAsSuite, setExportAsSuite }: Props) {
+export function Body({ code, exportAsProject, setExportAsProject }: Props) {
   return (
     <EuiFlyoutBody>
       <EuiCheckbox
-        id="export-as-suite-checkbox"
-        label="Export as suite"
-        checked={exportAsSuite}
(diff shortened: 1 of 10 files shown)


src/components/ExportScriptFlyout/Body.tsx
```diff
@@ -28,18 +28,18 @@ import type { Setter } from '../../common/types';
...
-export function Body({ code, exportAsSuite, setExportAsSuite }: Props) {
+export function Body({ code, exportAsProject, setExportAsProject }: Props) {
   return (
     <EuiFlyoutBody>
       <EuiCheckbox
-        id="export-as-suite-checkbox"
-        label="Export as suite"
-        checked={exportAsSuite}
-        onChange={() => setExportAsSuite(!exportAsSuite)}
+        id="export-as-project-checkbox"
```
justinkambic, line 39: TBH I'm unsure why this property is even here. Perhaps it's required by EUI. In any case, it's fine to rename as it's unreferenced elsewhere in the code.

justinkambic (approved):
LGTM

src/components/ExportScriptFlyout/Body.tsx
kyungeunni: Yes, id is required for EuiCheckbox component

PR #284, opened 2022-07-19
fix: always wait for url when navigation is signaled

justinkambic:
## Summary

Removes the comment that gets added before the navigation action is recorded. 

Record a journey that includes URL navigation. Make sure that your JS output will run properly.


justinkambic (commented):
LGTM, I tested this locally and the output was as I expected. But I can't approve my own PR even if it includes someone else's changes 😆

andrewvc (approved):
LGTM

Issue #263, opened 2022-06-01
Define a process for updating PW for each stack release


justinkambic:
Over time it will be necessary to upgrade Playwright versions regularly. This will be a result of a continuation of the work we accomplish in https://github.com/elastic/synthetics-recorder/issues/262.

kyungeunni:
We defined updating PW process in [README.md](https://github.com/elastic/synthetics-recorder#managing-playwright-dependency). It is focused on updating the PW version whenever Synthetics agent uses a newer version of PW, so updating PW regularly for each stack release should be in Synthetics agent's responsibility, then the recorder will follow its update. cc @vigneshshanmugam

PR #285, opened 2022-07-26
chore: update dependency @elastic/synthetics and Playwright

### LINKED ISSUE #262: Migrate Elastic Playwright fork to later version of official Playwright
The version of Playwright we reference today is very out of date. Before we can make the Recorder GA, we are going to need to update Elastic's Playwright fork to the latest version referenced by Synthetics. Once we do this, we unblock ourselves from upgrading Playwright, Elastic Synthetics, etc.

This is important from a feature perspective, as well as a security standpoint. We can't make the Recorder GA until we're able to respond to emergency patches to all of its dependencies, including Synthetics/Playwright. Without accomplishing this upgrade, we will not be able to quickly patch our software in the event of a serious security issue, and thus we're not a candidate for production use until we have that capability.

## TODO

- [x] Update `@elastic/playwright` to latest upstream:main, or fresh fork the repo
- [x] Apply the patch we have in `separate-recorder` branch
- [x] Use latest `@elastic/synthetics` in recorder

kyungeunni:
closes #262 and #263
relates to https://github.com/elastic/playwright/pull/1

## Summary
This PR updates a couple of dependencies, the latest [@elastic/synthetics](https://github.com/elastic/synthetics) agent (0.0.1-beta.31) and playwright version that the agent is using.
This update is especially exciting as Synthetics Recorder now generates scripts with [Locator API](https://playwright.dev/docs/api/class-locator) 🎉 

Some notable changes:
- Uses Node version to 16.15.0, npm >= 7 (package-lock.json is now version 2) 
- Removes `npm-force-resolutions` preinstall hook as it pollutes the npm install behaviour(since npm@7)
- Adds temporary type definition around `ActionInContext` and `Step` to make it compatible with the newer Playwright. These type definitions will be deleted once synthetics agent exports the compatible type definition again
- After updating Node version, we needed to build a new Docker image. With the new image and the new node, CI was having some issues with the permission when executing npm scripts. It seems like the docker tries to run the node executable installed by the host workspace that is mounted as volume. As a workaround, we install node dependencies rather than mounting the workspace volume when building the image.

Pull this branch and run `npm ci` and run `npm run dev`. Confirm that everything works as before, also note that the generated code now contains `page.locator(...)`.

## TODO
-  ~Find out what has changed to the signal url when recording the script. It's making the test fail as it waits for the navigation to the URL with the specific querystring which usually generated randomly on each navigation.~
**edit**: _The same behaviour is observed in the main branch as well so thus it is not related to the playwright version update. will address it separately

## Next Steps
- Test releases
- Delegate the code generation to Synthetics agent


package.json
```diff
@@ -91,20 +90,15 @@
...
-    "trim": "0.0.3",
-    "caniuse-lite": "1.0.30001359"
-  },
   "browserslist": [
     "last 1 version"
   ],
   "engines": {
-    "node": ">14.14.0",
-    "npm": "6"
+    "node": ">=16.15.0"
```
vigneshshanmugam, line 102: Do you want to force npm as well? Because otherwise we wont get package lock 2.0 IIRC. 

Also we should do this for Synthetics as well ASAP, or we do synthetics first and do the lock version switch to keep things simpler. I will leave it up to you.

scripts/install-pw.js
```diff
@@ -24,10 +24,10 @@ THE SOFTWARE.
 
 // Load the process.env config from .env file
 require('dotenv').config();
-
 const { Registry } = require('playwright/lib/utils/registry');
-const SYNTHETICS_BROWSER_REVISIONS = require('@elastic/synthetics/node_modules/playwright-chromium/browsers.json');
-
+const PLAYWRIGHT_CORE_PATH = require.resolve('playwright-core');
+const BROWSER_PATH = PLAYWRIGHT_CORE_PATH.replace('index.js', 'browsers.json');
```
vigneshshanmugam, line 29: The reason why we want to extract the browser version from Synthetics agent is because we want the recorder to record and run results in the same exact browser version. 

PW browser.json depends on why we updated the fork and might not match the Synthetics test version we want to use.

vigneshshanmugam: So it needs to be `@elastic/synthetics/node_modules/playwright-core/browsers.json`

README.md
```diff
@@ -84,51 +58,45 @@ Run the recorder app in dev mode.
...
-in scope for all its references.
-
-```javascript
-step('Click text=Babel Minify', async () => {
-  // now all references to `page2` occur while it is still in scope
-  const [page2] = await Promise.all([page.waitForEvent('popup'), page.click('text=Babel Minify')]);
-  await page2.click(':nth-match(a:has-text("babel-minify"), 3)');
-  await page2.close();
-});
+2. Confirm the Playwright version from Synthetics agent's `package.json`, we will refer it as `v${NEW_VERSION}`. Checkout to `NEW_VERSION` tag, and create a new branch from there:
```
vigneshshanmugam, line 74: This is confusing and assumes Synthetics agent PW version with the branch version which is not we are doing with the rebasing. Instead it would be the latest version of Microsoft playwright that was released. 

I do understand the intention here that we can keep a versioning scheme, But then we would have to assume the synthetics agent is always updated to the PW latest version.

vigneshshanmugam (commented):
Did a first pass of the code, Looks great 🎉 

I will try to run and test locally.

package.json
kyungeunni: My naive intention here was that node 16.15.0 comes with npm 8 so we could skip enforcing the npm version, but it is always better to be explicit 👍 Will update it.

scripts/install-pw.js
kyungeunni: What I've seen with the newer npm is that it downloads `playwright-core` to top-level `node_modules` rather than `@elastic/synthetics/node_modules` (`@elastic/synthetics/node_modules` doesn't have `playwright-core`):

image
The forked PW is downloaded and aliased as `playwright` so I think we still are importing the `browser.json` from the `playwright-core` package that is used by Synthetics agent.

README.md
kyungeunni: We might need to sync on this, but maybe we have a different approach on this.

Are you saying that the recorder should use the latest PW (let's say `1.24` or even higher that is not released) whereas the synthetics agent uses an older version(1.20.1 as of today)? I thought the intention was to use the same version of the playwright on both sides so we don't have any surprises. 

For this PR what I did was:
1. fetch `main` from upstream (PW latest version is 1.25.0-next)
2. create a dev branch that is sync up with the main and an extra commit with the modification for the recorder
3. Synthetics agent is using 1.20.1, so I checkout to `v1.20.1`, create a new branch named `1.20.1-recorder`, and then cherry-pick the commit from the dev branch
4. Use `1.20.1-recorder` branch from the recorder

But from my understanding, you are suggesting we do:
1. fetch `main` from upstream (PW latest version is 1.25.0-next)
2. create a dev branch that is sync up with the main and an extra commit with the modification for the recorder
3. Use dev branch from the recorder

package.json
vigneshshanmugam: Thanks.

scripts/install-pw.js
vigneshshanmugam: Ahh, npm 7 has workspaces set by default. Good callout, can you add a comment for the same - Its hard to grok in.

README.md
vigneshshanmugam: Myself and Kyungeun Synced on this separately, It was a misunderstanding in my part. The outcome is to update the docs with bit more detail so we can give more clarity around what versioning scheme we want to follow.

scripts/install-pw.js
kyungeunni: Sure, added!!

README.md
kyungeunni: I've updated the doc, let me know what you think! 🙏

kyungeunni: I've updated the doc, let me know what you think! 🙏

PR #287, opened 2022-08-10
fix: allow editing value when editing `fill` action

### LINKED ISSUE #282: [Bug] Cannot edit fill value on win11 after recording
### Bug summary

Editing an input fill command is not possible (at least not on my win11 box).

### Recorder Version

1.0.3

### Steps to reproduce

Recorded a journey, visiting google.com and typing some search text in the box. Later, could not edit the fill value in the UI.

image

### Expected behavior

_No response_

### Additional information

_No response_

kyungeunni:
fixes #282 

Allow editing `Value` field for the fill command.

#### How to validate the change
Record a journey that types some text in an input element. 
Locate the `fill` action and go to Edit action. Check that the Value input field is editable and the change is applied when clicking the `Save` button.

### DIFF
--- a/src/components/ActionDetail/ActionDetail.test.tsx
+++ b/src/components/ActionDetail/ActionDetail.test.tsx
@@ -22,20 +22,25 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 */
 import React from 'react';
-import { fireEvent, getByRole } from '@testing-library/react';
+import { fireEvent } from '@testing-library/react';
 import { createAction } from '../../../common/helper/test/createAction';
 import { render } from '../../helpers/test';
 import { ActionDetail } from './ActionDetail';
+import { ActionContext } from '../../../common/types';
+import { IStepsContext, StepsContext } from '../../contexts/StepsContext';
+import { getStepsContextDefaults } from '../../helpers/test/defaults';
 
 describe('<ActionDetail />', () => {
   afterEach(() => {
-    // restore the spy created with spyOn
     jest.restoreAllMocks();
   });
 
   describe('navigate', () => {
     const url = 'https://example.com';
-    const navigateAction = createAction('navigate', { action: { url } });
+    let navigateAction: ActionContext;
+    beforeEach(() => {
+      navigateAction = createAction('navigate', { action: { url } });
+    });
 
     it('has url input', () => {
       const { getByLabelText } = render(
@@ -47,16 +52,12 @@ describe('<ActionDetail />', () => {
     });
 
     it('updates url', () => {
-      const update = jest.fn(updatedAction => {
+      const onUpdateActionMock = jest.fn(updatedAction => {
         return updatedAction;
       });
-      const { getByLabelText, getByRole } = render(
-        <ActionDetail
-          actionContext={navigateAction}
-          actionIndex={0}
-          stepIndex={0}
-          update={update}
-        />
+      const { getByLabelText, getByRole } = renderWithStepsContext(
+        <ActionDetail actionContext={navigateAction} actionIndex={0} stepIndex={0} />,
+        { onUpdateAction: onUpdateActionMock }
       );
       const input = getByLabelText(/navigate/i);
 
@@ -64,16 +65,19 @@ describe('<ActionDetail />', () => {
       fireEvent.change(input, { target: { value: newUrl } });
       const button = getByRole('button', { name: /save/i });
       fireEvent.click(button);
-      expect(update).toHaveBeenCalled();
+      expect(onUpdateActionMock).toHaveBeenCalled();
       const {
         value: { action },
-      } = update.mock.results[0];
+      } = onUpdateActionMock.mock.results[0];
       expect(action.url).toEqual(newUrl);
     });
   });
 
   describe('click', () => {
-    const clickAction = createAction('click', { action: { selector: '#elem-id', url: '' } });
+    let clickAction: ActionContext;
+    beforeEach(() => {
+      clickAction = createAction('click', { action: { selector: '#elem-id', url: '' } });
+    });
 
     it('has selector value in input', () => {
       const { getByLabelText } = render(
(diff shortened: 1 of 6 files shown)


src/components/ActionDetail/ActionDetail.test.tsx
```diff
@@ -0,0 +1,147 @@
...
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+import React from 'react';
+import { fireEvent, getByRole } from '@testing-library/react';
```
justinkambic, line 25:
```suggestion
import { fireEvent } from '@testing-library/react';
```

src/components/ActionDetail/ActionDetail.test.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+describe('<ActionDetail />', () => {
+  afterEach(() => {
+    // restore the spy created with spyOn
+    jest.restoreAllMocks();
+  });
+
+  describe('navigate', () => {
+    const url = 'https://example.com';
+    const navigateAction = createAction('navigate', { action: { url } });
```
justinkambic, line 38: I find from a maintenance perspective, re-initializing vars like this for each test tends to avoid unwanted side effects that can break tests and be difficult to debug .

```suggestion
    let navigateAction: ActionContext;
    beforeEach(() => {
      navigateAction = createAction('navigate', { action: { url } });
    });
```

src/components/ActionDetail/ActionDetail.test.tsx
```diff
@@ -0,0 +1,147 @@
...
+      expect(update).toHaveBeenCalled();
+      const {
+        value: { action },
+      } = update.mock.results[0];
+      expect(action.url).toEqual(newUrl);
+    });
+  });
+
+  describe('click', () => {
+    const clickAction = createAction('click', { action: { selector: '#elem-id', url: '' } });
```
justinkambic, line 76: Same recommendation as above for `beforeEach`.

src/components/ActionDetail/ActionDetail.test.tsx
```diff
@@ -0,0 +1,147 @@
...
+      expect(update).toHaveBeenCalled();
+      const {
+        value: { action },
+      } = update.mock.results[0];
+      expect(action.selector).toEqual('#new-elem-id');
+    });
+  });
+
+  describe('fill', () => {
+    const fillAction = createAction('fill', {
```
justinkambic, line 108: Same recommendation for `beforeEach`.

src/components/ActionDetail/ActionDetail.tsx
```diff
@@ -37,50 +37,59 @@ import { FormControl } from './FormControl';
...
+  url: string;
+  selector: string;
+  text?: string;
+}
 interface IActionDetail {
   actionContext: ActionContext;
   actionIndex: number;
   stepIndex: number;
   close?: () => void;
+  update?: (action: ActionContext, stepIndex: number, actionIndex: number) => void;
```
justinkambic, line 66: Where is this prop actually supplied to `ActionDetail`? It seems the only place where we're applying this prop is in the tests, unless I'm missing something?

src/components/ActionDetail/ActionDetail.tsx
```diff
@@ -127,15 +136,7 @@ export function ActionDetail({ actionContext, actionIndex, close, stepIndex }: I
...
-            onClick={() => {
-              if (url) {
-                onURLChange(url);
-              } else if (selector && !action.isAssert) {
-                onSelectorChange(selector);
-              } else if (text) {
-                onTextChange(text);
-              }
-            }}
+            onClick={() => updateAction()}
```
justinkambic, line 139:
```suggestion
            onClick={updateAction}
```

justinkambic (commented):
Had a few recommendations, let me know what you think. Thanks for writing such a great test file to accompany the changes 💯.

src/components/ActionDetail/ActionDetail.test.tsx
kyungeunni: Yes that's a good point!! will update it as you suggested 👍

src/components/ActionDetail/ActionDetail.tsx
kyungeunni: Yes, Actually this is the part I wasn't really happy with.. I tried to spy on the `onUpdateAction` but didn't want to introduce the `StepContext` for the unit tests so I ended up adding the prop that is only used by tests. Do you have any suggestions?

kyungeunni: Actually, I've updated the tests to use StepContext so that we didn't need to add a new prop `update` just for the test purpose. Thanks for bringing this to my attention.

src/components/ActionDetail/ActionDetail.tsx
```diff
@@ -37,50 +37,59 @@ import { FormControl } from './FormControl';
...
-  const onSelectorChange = (value: string) => {
-    if (!value) return;
-    setSelector(value);
-    onUpdateAction(createUpdatedAction('selector', value, actionContext), stepIndex, actionIndex);
-  };
-  const onTextChange = (value: string) => {
-    if (!value) return;
-    setText(value);
-    onUpdateAction(createUpdatedAction('text', value, actionContext), stepIndex, actionIndex);
+  const updateAction = () => {
```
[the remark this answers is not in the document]
kyungeunni: Thanks for the suggestion 👍 Although I thought there is a minimal chance that we'd leverage the full advantage of memorized function as there are so many deps participating there. (also as you mentioned the render is quite cheap).

src/components/ActionDetail/ActionDetail.tsx
justinkambic: A common pattern, too, is to overload the RTL `render` function with a custom test helper that will wrap the components with some custom boilerplate and remove that burden from the test files.

I think we likely even have one of those helpers in this project already. I can check for you in a bit.

src/components/ActionDetail/ActionDetail.test.tsx
```diff
@@ -0,0 +1,161 @@
...
+      const {
+        value: { action },
+      } = onUpdateActionMock.mock.results[0];
+      expect(action.selector).toEqual('#new-elem-id');
+      expect(action.text).toEqual('Elastic');
+    });
+  });
+});
+
+function renderWithStepsContext(childComp: React.ReactChild, overrides?: Partial<IStepsContext>) {
```
justinkambic, line 158: I think we don't need this, have a look at [this helper function](https://github.com/elastic/synthetics-recorder/blob/acd37f14df07399cfdd4a96b2bb6694f5796c9db/src/helpers/test/render.tsx#L40).

kyungeunni: Oh nice!

kyungeunni: I've updated it using the helper function 👍

kyungeunni:
Merged it to include this bug fix in a new release. I will revisit the last changes I made around tests if we are not happy about them(which has been iterated a few times).

PR #288, opened 2022-08-17
chore: update electron version

kyungeunni:
Brings electron to up-to-date version. It also mitigates one of the security vulnerability issues.

### DIFF
--- a/electron/electron.ts
+++ b/electron/electron.ts
@@ -53,7 +53,6 @@ async function createWindow() {
       devTools: isDev || IS_TEST,
       nodeIntegration: true,
       contextIsolation: false,
-      nativeWindowOpen: true,
     },
   });
 
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "@types/styled-components": "^5.1.25",
     "ajv": "^8.11.0",
     "concurrently": "^7.2.2",
-    "electron": "^14.2.9",
+    "electron": "^20.0.2",
     "electron-builder": "^23.3.1",
     "eslint": "^8.19.0",
     "eslint-config-prettier": "^8.5.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -35,7 +35,7 @@
         "@types/styled-components": "^5.1.25",
         "ajv": "^8.11.0",
         "concurrently": "^7.2.2",
-        "electron": "^14.2.9",
+        "electron": "^20.0.2",
         "electron-builder": "^23.3.1",
         "eslint": "^8.19.0",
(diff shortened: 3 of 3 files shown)


electron/electron.ts
```diff
@@ -53,7 +53,6 @@ async function createWindow() {
       devTools: isDev || IS_TEST,
       nodeIntegration: true,
       contextIsolation: false,
-      nativeWindowOpen: true,
```
kyungeunni, line 56: Now it's on by default: https://www.electronjs.org/docs/latest/breaking-changes#removed-nativewindowopen

justinkambic (approved):
LGTM!

package.json
```diff
@@ -57,7 +57,7 @@
     "@types/styled-components": "^5.1.25",
     "ajv": "^8.11.0",
     "concurrently": "^7.2.2",
-    "electron": "^14.2.9",
+    "electron": "^20.0.2",
```
vigneshshanmugam, line 60: wow thats a huge bump. Is everything working properly? The sandboxing and context isolation comes to mind tha always breaks with major version bumps.

justinkambic, line 60: I tested this earlier and everything seemed ok. Per @vigneshshanmugam's comment I went back and checked the Test functionality again and now I'm seeing an error like:

```
[2] Error: 
[2] Something went wrong installing the "sharp" module
[2] 
[2] Cannot find module '../build/Release/sharp-darwin-arm64v8.node'
```

The Test feature is broken on this branch.

justinkambic (changes requested):
Sorry, I might not have re-bootstrapped when I tested this branch before, I'm seeing errors now.

package.json
kyungeunni: It might be that you have installed the modules while you were on node 14? Node 14 doesn't support Apple Silicon so when installing sharp(which is a native module), it downloads darwin-x64. But the latest version of electron uses node 16 and it can recognize if the host machine is Apple Silicon and it expects to find the arm build of sharp.

Try to run:
1. nvm use
2. npm ci

and run the test feature again, you shouldn't see the warning again.

kyungeunni: I've double-checked, and manual testing and the e2e tests all seem to be ok.

justinkambic: You are probably right. I likely tested in a shell with the proper node version firstly, and the second time failed to use the right version. I'll test again soon.

justinkambic: Yeah you were right, I must've not been on the appropriate node version.

justinkambic (approved):
Features seem to be working alright for me, when using the correct node version.

PR #290, opened 2022-08-24
chore: notarize Mac applications

kyungeunni:
closes #51 

## Summary
Notarize Mac applications so that the users can use the app without worrying about malicious code.

This PR introduces [electron-notarize](https://github.com/electron/electron-notarize) module to notarize the app and it is used when the Electron builder's `afterSign` hook is called. It is a recommended by electron-builder maintainers.

## How to validate the change
After the artefacts are released, download them, and run the following commands to verify the app has been notarized.

(Artifacts generated from this branch can be found [here](https://github.com/elastic/synthetics-recorder/releases/tag/untagged-053cedd0e62b7b3b9d3f))

- Validate the app is notarized(should see _source=**Notarized Developer ID**_):
```
❯ spctl -a -t exec -vv $PATH_TO_APP/Elastic Synthetics Recorder.app
$PATH_TO_APP//Elastic Synthetics Recorder.app: accepted
source=Notarized Developer ID
origin=Developer ID Application: Elasticsearch, Inc (2BT3HPN62Z)
```

- Check the ticket is stapled(should see **_The validate action worked!_**):
```
❯ xcrun stapler validate $PATH_TO_APP/Elastic Synthetics Recorder.app
Processing: $PATH_TO_APP/Elastic Synthetics Recorder.app
The validate action worked!
```
- When opening the app, the warning message _Elastic Synthetics Recorder" can't be opened because Apple cannot check it for malicious software_ shouldn't be shown, instead, the message could be something like this:
image

#### Checklist
I've check above points for all the artifacts below
- [x] arm64.dmg (after installing)
- [x] arm64.zip
- [x] x64.dmg (after installing)
- [x] x64.zip

PR #292, opened 2022-09-28
fix: use default entitlements

kyungeunni:
## Summary

There has been a case some of the functionalities don't work since #290 is merged
The only part that can affect this seems to be updated entitlements.
I will revert back to the previous setting which uses the default entitlements list provided by [electron-osx-sign](https://github.com/electron/osx-sign/pull/253)

Build the apps from this branch and test all of the mac artifacts, then verify`Test` feature works


kyungeunni:
Closing as entitlements are not an issue, and the default entitlements require extra permissions than we need.

PR #309, opened 2022-11-03
docker: use node docker images, mount volume and use node user

v1v:
## Summary

Revert partially https://github.com/elastic/synthetics-recorder/commit/5c91e170eded3f950443529bfdd7fe7f738100cc so it can mount the volume

## Implementation details

There are some issues while using the docker build with a docker copy entry since the workspace is stored hence it can cause other issues related to what's the workspace used for testing.

By reverting https://github.com/elastic/synthetics-recorder/commit/5c91e170eded3f950443529bfdd7fe7f738100cc , restoring the workspace, setting the uid and using the node docker images things should be back to normal.

It runs successfully in the CI

or 

```bash
$ .ci/scripts/docker-build.sh
$ .ci/scripts/run-test-in-docker.sh
```

### DIFF
--- a/.ci/scripts/docker-build.sh
+++ b/.ci/scripts/docker-build.sh
@@ -14,4 +14,5 @@ docker \
   --file e2e/Dockerfile.jenkins \
   --build-arg DISPLAY=":99" \
   --build-arg NODE_VERSION="${NODE_VERSION}" \
+  --build-arg UID="$UID" \
   --tag docker.elastic.co/observability-ci/synthetics-recorder:$TAG .
--- a/.ci/scripts/docker-build.sh
+++ b/.ci/scripts/docker-build.sh
@@ -3,8 +3,12 @@ set -eox pipefail
 
 TAG=${BRANCH_NAME:-'latest'}
 
+# Read node version from .nvmrc and remove v prefix.
+NODE_VERSION=$(cat .nvmrc | sed "s#v##g")
+
 docker \
   build \
   --file e2e/Dockerfile.jenkins \
   --build-arg DISPLAY=":99" \
+  --build-arg NODE_VERSION="${NODE_VERSION}" \
   --tag docker.elastic.co/observability-ci/synthetics-recorder:$TAG .
--- a/.ci/scripts/docker-build.sh
+++ b/.ci/scripts/docker-build.sh
@@ -6,6 +6,9 @@ TAG=${BRANCH_NAME:-'latest'}
 # Read node version from .nvmrc and remove v prefix.
 NODE_VERSION=$(cat .nvmrc | sed "s#v##g")
 
+# Download the docker image if possible
+docker pull docker.elastic.co/observability-ci/synthetics-recorder:$TAG || true
+
 docker \
   build \
   --file e2e/Dockerfile.jenkins \
--- a/.ci/scripts/run-test-in-docker.sh
+++ b/.ci/scripts/run-test-in-docker.sh
@@ -23,6 +23,7 @@ fi
 docker run \
   $DOCKER_RUN_OPTIONS \
   -u '0:0' \
+  -v "$(pwd):/synthetics-recorder" \
   -e NPM_COMMAND=${1:-''} \
   $DOCKER_IMAGE \
   .ci/scripts/run-test.sh
--- a/.ci/scripts/run-test-in-docker.sh
+++ b/.ci/scripts/run-test-in-docker.sh
@@ -22,7 +22,6 @@ fi
 # shellcheck disable=SC2086
 docker run \
   $DOCKER_RUN_OPTIONS \
-  --user node \
   -v "$(pwd):/synthetics-recorder" \
   -e NPM_COMMAND=${1:-'ci'} \
   $DOCKER_IMAGE \
--- a/.ci/scripts/run-test-in-docker.sh
+++ b/.ci/scripts/run-test-in-docker.sh
@@ -24,6 +24,6 @@ docker run \
   $DOCKER_RUN_OPTIONS \
   -u '0:0' \
   -v "$(pwd):/synthetics-recorder" \
-  -e NPM_COMMAND=${1:-''} \
+  -e NPM_COMMAND=${1:-'ci'} \
   $DOCKER_IMAGE \
   .ci/scripts/run-test.sh
--- a/.ci/scripts/run-test-in-docker.sh
+++ b/.ci/scripts/run-test-in-docker.sh
@@ -22,6 +22,8 @@ fi
 # shellcheck disable=SC2086
 docker run \
   $DOCKER_RUN_OPTIONS \
+  --user node \
+  -v ~/.npmrc:/home/node/.npmrc \
   -v "$(pwd):/synthetics-recorder" \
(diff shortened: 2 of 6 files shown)


.ci/scripts/run-test-in-docker.sh
```diff
@@ -23,6 +23,7 @@ fi
 docker run \
   $DOCKER_RUN_OPTIONS \
   -u '0:0' \
-  -e NPM_COMMAND=${1:-''} \
+  -v "$(pwd):/synthetics-recorder" \
+  -e NPM_COMMAND=${1:-'ci'} \
```
v1v, line 27: This one was causing the `npm ci` to be skipped since no params were passed and hence `''` was the default value

e2e/Dockerfile.jenkins
```diff
@@ -29,13 +29,12 @@ RUN apt-get -y -qq install xvfb \
...
-# ADD .nvmrc /synthetics-recorder/.nvmrc
+ADD .nvmrc /synthetics-recorder/.nvmrc
 
 # Installing NVM
 ENV NVM_DIR /root
 RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.0/install.sh | bash
 RUN source "${NVM_DIR}/nvm.sh" --install \
-  && nvm use && npm ci
\ No newline at end of file
+  && nvm install $(cat .nvmrc) && nvm use $(cat .nvmrc)
```
v1v, line 40: I don't know if the issue was regarding `.nvmrc` with wrong permissions, but if that's not the case, then

```suggestion
  && nvm use
```

e2e/Dockerfile.jenkins
```diff
@@ -1,7 +1,5 @@
-FROM ubuntu:20.04
-
-# Setting bash as the default shell
-RUN rm /bin/sh && ln -s /bin/bash /bin/sh
+ARG NODE_VERSION=16.15.0
+FROM node:$NODE_VERSION-buster
```
v1v, line 2: it's even possible to avoid `-buster` see https://hub.docker.com/layers/library/node/16.15.0/images/sha256-2d085d25d573a58266319362f32e978f94e06214a62fae15c02d4c4f0567b64f?context=explore


```suggestion
FROM node:$NODE_VERSION
```

v1v:
```
[2022-11-03T18:49:42.694Z] npm ERR! sharp: Please see https://sharp.pixelplumbing.com/install for required dependencies
[2022-11-03T18:49:42.694Z] npm ERR! sharp: Installation error: EACCES: permission denied, mkdir '/root/.npm'
```


https://github.com/elastic/synthetics-recorder/pull/309/commits/2a89d6006bdd123b369540bd139f45f030538df5 might help

> If you need to install global npm dependencies, it is recommended to place those dependencies in the [non-root user](https://github.com/nodejs/docker-node/blob/main/docs/BestPractices.md#non-root-user) directory. To achieve this, add the following line to your Dockerfile

```
ENV NPM_CONFIG_PREFIX=/home/node/.npm-global
ENV PATH=$PATH:/home/node/.npm-global/bin # optionally if you want to run npm global bin without specifying path
```

.ci/scripts/run-test-in-docker.sh
kyungeunni: AFAIR `npm ci` was already run from one of the previous steps so we didn't need to run it again

v1v: That's correct, I'll remove it then

.ci/scripts/docker-build.sh
```diff
@@ -3,8 +3,16 @@ set -eox pipefail
...
+
+# Download the docker image if possible
+docker pull docker.elastic.co/observability-ci/synthetics-recorder:$TAG || true
+
 docker \
   build \
   --file e2e/Dockerfile.jenkins \
   --build-arg DISPLAY=":99" \
+  --build-arg NODE_VERSION="${NODE_VERSION}" \
+  --build-arg UID="$UID" \
```
v1v, line 17: Ensure the UID is populated to the `node` user so when mounting the volume the docker container can write in the directory. Docker volumes in MacOS and Linux differ a bit, so I could not reproduce the issue locally since I use MacOS but it failed in Linux unless resetting the UID for the node user

.ci/scripts/run-test-in-docker.sh
```diff
@@ -22,7 +22,7 @@ fi
 # shellcheck disable=SC2086
 docker run \
   $DOCKER_RUN_OPTIONS \
-  -u '0:0' \
```
v1v, line 25: Unrequired while running the tests

.gitignore
```diff
@@ -13,3 +13,5 @@ eslint-junit.xml
 *-junit.xml
 .DS_Store
 tsconfig.tsbuildinfo
+
+.npm
```
v1v, line 17:
```suggestion
```

e2e/Dockerfile.jenkins
```diff
@@ -1,7 +1,5 @@
-FROM ubuntu:20.04
-
-# Setting bash as the default shell
-RUN rm /bin/sh && ln -s /bin/bash /bin/sh
+ARG NODE_VERSION=16.15.0
```
v1v, line 1: A simpler Dockerfile using the official node docker images, rather than nvm, the NODE_VERSION is set as part of the `docker-build.sh` by reading the version set in the `.npmrc`

kyungeunni: I love that we are removing nvm! 👍 What do you think about not providing the default value for `NODE_VERSION` here? I imagined, in a few months, someone might think that we need to update node version in both `.nvmrc` and this file If they don't know `docker-build.sh` is reading the version set in the `.npmrc`

.ci/scripts/docker-build.sh
kyungeunni: Thanks for the info!!

e2e/Dockerfile.jenkins
```diff
@@ -29,13 +27,10 @@ RUN apt-get -y -qq install xvfb \
...
-COPY . ./synthetics-recorder
 WORKDIR /synthetics-recorder
-# Copy .nvmrc as long as you run the script in .ci/scripts/docker-build.sh
-# ADD .nvmrc /synthetics-recorder/.nvmrc
 
-# Installing NVM
-ENV NVM_DIR /root
-RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.0/install.sh | bash
-RUN source "${NVM_DIR}/nvm.sh" --install \
-  && nvm use && npm ci
```
kyungeunni, line 41: Love the changes made here ✨

kyungeunni (approved):
Thank you so much, @v1v! this looks great!! ✨

e2e/Dockerfile.jenkins
v1v: The idea of build args for Docker is to override them when running `docker build`. If it causes any issues, we can remove it, no strong opinion.

v1v:
```suggestion
ARG NODE_VERSION
```

PR #337, opened 2022-12-23
fix: import correct module for notarizing

kyungeunni:
Recently updated `electron-notarize` module to recommended `@electron/notarize`, but forgot to update the module name where it's used.

Also fixed one of the functions that has been moved to a different path.


kyungeunni:
Check failed due to https://github.com/elastic/synthetics-recorder/issues/338

PR #339, opened 2022-12-23
Fix permission denied in CI

reakaleek:
## Summary

https://github.com/elastic/synthetics-recorder/issues/338 Fix permission denied during junit test recording.

## Implementation details

The xml files generated by the junit test reporter were created in the docker container with the root user. Hence, Jenkins was not able to read the file to execute the junit function.

Reusing the `UID` and `GID` of Jenkins in the docker container resolves the issue.

Further, the `package-lock.json` was updated with the command `npm update` because of the following error (inspired by https://github.com/npm/cli/issues/3374#issuecomment-1188070482):

image

```
npm ERR! Cannot read properties of null (reading 'pickAlgorithm')
```
> Source: https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-rum%2Fsynthetics-recorder-mbp/detail/PR-339/4/pipeline/86

during `npm ci`. This behavior happend both on Jenkins and locally.

Running `npm clear cache --force` did not help.

## How to validate this change

No permission errors during the Jenkins build.

Issue #338, opened 2022-12-23
[Bug] Test fails due to permission issue in CI


kyungeunni:
### Bug summary

When running test in CI, it fails as the file read or write can't be perfromed due to permission issues.
Most recent example can be found [here](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-rum%2Fsynthetics-recorder-mbp/detail/PR-337/1/pipeline#step-277-log-214) and here's the screenshot
<img width="810" alt="image" src="https://user-images.githubusercontent.com/16660866/209285324-7afd1a47-ab66-4009-8309-f01bf71717be.png">

cc @elastic/observablt-robots @v1v 

### Recorder Version

N/A

### Steps to reproduce

Run tests

kuisathaverat:
@reakaleek check if the test run in a Docker container that uses the root user, if so, change it to use the local user with `-u $(id -u):$(id -g)`

reakaleek:
Thank you for the tip. I had the same idea!

Pretty sure this is the issue, the tests ur running in a docker container, just as you presumed.

However, I'm currently struggling with another issue.

While trying to reproduce it, my [PR](https://github.com/elastic/synthetics-recorder/pull/339) already fails at the install step.

```
pm ERR! Cannot read properties of null (reading 'pickAlgorithm')
```

I was able to get the same error locally.

I tried multiple things like

`npm cache clear --force` or `npm install` instead of `npm ci`.

The only approach that worked is `npm update` but this updates all the packages in the `package-lock.json` file.

I will do it for now so I can continue..

reakaleek:
@kuisathaverat Thank you for the tip. I had the same idea!

Pretty sure this is the issue, the tests ur running in a docker container, just as you presumed.

However, I'm currently struggling with another issue.

While trying to reproduce it, my [PR](https://github.com/elastic/synthetics-recorder/pull/339) already fails at the install step.

```
pm ERR! Cannot read properties of null (reading 'pickAlgorithm')
```

I was able to get the same error locally.

I tried multiple things like

`npm cache clear --force` or `npm install` instead of `npm ci`.

The only approach that worked is `npm update` but this updates all the packages in the `package-lock.json` file.

I will do it for now so I can continue..

kyungeunni:
Closing this as we moved lint & test step to github actions. I will create a follow-up ticket for the remaining [tasks](https://github.com/elastic/synthetics-recorder/pull/341#issuecomment-1372402701).

PR #181, opened 2022-03-30
ci: release tag based (stage 1)

v1v:
## Summary

Automate the release and publishing steps.

## Implementation details

* Tag based event (`v<major>.<minor>.*`)
* Notify if something bad happened
* Create draft release
* Skip all the stages but build and release when a tag release is created, every commit is validated in the CI either coming from a PR or merged to the main branch or release branches, therefore there is no need to validate it again when a tag is created from a validated commit.

## Next steps

* Package and upload the artifacts as assets to the GitHub release
* Sign the artifacts

## Requirements
* Confirm what slack/email to be used for notifications.

PR #341, opened 2023-01-04
test: try running tests on github action

kyungeunni:
## Summary
Use github actions for `Test` step to simplify the process and reduce the maintenance burden

## Implementation details
We have had many issues with the testing environment in the past year, often caused by npm changing its behaviour or when using newer node versions (some contexts can be found [here](https://github.com/elastic/observability-robots/issues/1453) )
Having the permission issue #338 recently, and the possible fix #339 causing another issue affecting test results, I wanted to explore the github actions for running tests. I've tried to run the tests on macos(this let us bypass the setup for `Xvfb`) and It seems quite promising. Here's the impact I've found when opting in for github actions:

### Pros:
- Do not need to build the docker image and maintain it. The docker image is only used by the test step and when we bump the release step, it doesn't run the test step anyways so it won't affect anything else.
- It can cache the npm modules until `package-lock.json` changes) so it _can_ reduce the time for checks significantly
- We can easily reproduce the same behavior as it does on my local machine. previously tests didn't fail at all on CI but failed on my local machine. Running tests locally using the docker image doesn't really work well for this project, as we use a native module(sharp) and the developers use macOS, it throws as the module is not compatible and we might need force-install the native module built for linux.(yet there are some differences even though we do that as docker for Mac behaves differently to linux)

### Cons:
- Won't have CI summary comments created by Jenkins

## Next Steps
If we are happy to go forward with github actions, we need to rearrange the Jenkins file and clean-up related scripts, it will be used mainly for the release pipelines(currently release pipeline is triggered when a tag is pushed). 

push to this branch

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -104,47 +104,6 @@ pipeline {
         }
       }
     }
-    stage('Docker build') {
-      options { skipDefaultCheckout() }
-      when {
-        beforeAgent true
-        anyOf {
-          expression { env.DOCKER_CHANGES == 'true' }
-          expression { return env.GITHUB_COMMENT?.contains('all') }
-        }
-      }
-      steps {
-        withGithubNotify(context: 'Docker Build') {
-          dir("${BASE_DIR}"){
-            dockerLogin(secret: "${DOCKER_ELASTIC_SECRET}", registry: "${DOCKER_REGISTRY}")
-            sh(label: 'docker build', script: ".ci/scripts/docker-build.sh")
-            retryWithSleep(retries: 3, seconds: 5, backoff: true){
-              sh(label: 'docker push', script: ".ci/scripts/docker-push.sh")
-            }
-          }
-        }
-      }
-    }
-    stage('Test'){
-      options { skipDefaultCheckout() }
-      when {
-        beforeAgent true
-        expression { return !isTag() }
-      }
-      steps {
-        withGithubNotify(context: "Test") {
-          dir("${BASE_DIR}"){
-            // NOTE: It uses the current workspace that was configured with npm ci
-            sh(label: 'run tests', script: '.ci/scripts/run-test-in-docker.sh')
-          }
-        }
-      }
-      post {
-        always {
-          junit(allowEmptyResults: true, keepLongStdio: true, testResults: "${BASE_DIR}/*junit.xml")
-        }
-      }
-    }
     stage('Release') {
       options { skipDefaultCheckout() }
       when {
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -43,67 +43,6 @@ pipeline {
         }
       }
     }
-    stage('Install') {
-      options { skipDefaultCheckout() }
-      when {
-        beforeAgent true
-        expression { return !isTag() }
-      }
-      steps {
-        withGithubNotify(context: "Install") {
-          withNodeJSEnv() {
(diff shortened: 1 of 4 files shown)


kyungeunni:
For some reason, gh action is not triggered

reakaleek:
> For some reason, gh action is not triggered

Yes, this is an unfortunate behaviour. When the repository never had a GH actions workflow before. 
We usually added a boilerplate workflow to initialize it.

Will create one, and let's try to get it merged.

reakaleek:
Glad it works now!

Yes, when migrating to GH Actions we want to keep all behaviours the same if possible.

- *Test Summary*: For junit test summary report already have an established pattern that works with fork PRs in mind.
    - See https://github.com/elastic/gmux/tree/main/.github/workflows as an example, where the `test-reporter.yml` workflow is triggered by the `test.yml` workflow.
    - Keep in mind this will only work if `test-reporter.yml` is already merged into `main`
- *Release Flow*: As far as I can see, you only migrated the test part. This will cause the tests to not run when performing a release.
    - We should also aim for migrating this part too. Unfortunately, we actually don't have an established "pattern" for this yet and I'm actually currently trying to design/poc this flow in GH Actions myself.
- *Alerting/Notifications*: If there are slack notifications, we definitely want to migrate those too. Any stakeholder might be surprised they are gone otherwise.

If we are okay with missing this functionality temporarily you can go ahead with this one and only apply what is important for you. As I'm currently working mainly on Jenkins to GHActions migrations, I could prioritize this repo and implement/migrate the missing pieces.

**Note**: Looking at my migration spreadsheet I can see the migration for this repo is actually undecided yet (according to the sheet)

@cachedout / @elastic/observablt-ci  WDYT?

kyungeunni:
Thanks, @reakaleek!
just one comment on this part:

> Release Flow: As far as I can see, you only migrated the test part. This will cause the tests to not run when performing a release.

We are not running tests when releasing even now(see comment [here](https://github.com/elastic/synthetics-recorder/pull/181#issue-1186688938)), so this is not a new behavior.

reakaleek (approved):

.github/workflows/test.yml
```diff
@@ -0,0 +1,30 @@
...
+      - '**.md'
+
+jobs:
+  test:
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16.15.0
```
v1v, line 19: What do you think if using the [node-version-file](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file) instead a fixed version instead?

```suggestion
          node-version-file: '.nvmrc'
```

kyungeunni: love that, thanks @v1v

.ci/Jenkinsfile
```diff
@@ -43,108 +43,6 @@ pipeline {
         }
       }
     }
-    stage('Install') {
```
kyungeunni, line 46: Removing Install, build, and lint stages too, as `lint` is moved to test.yml and Release stage does `npm ci` and `npm run build` within its stage.

kyungeunni:
Also I've double-checked that the release CI is unaffected and successfully release the apps.

justinkambic:
Skipping Post-FF Testing as this feature is integral to the merge process and if there were problems we'd see/remediate them as part of normal day-to-day workflow.

PR #343, opened 2023-01-04
chore: call click stop recording to avoid tests to hang

devcorpio:
# Context

The execution of the e2e from dev machines (tested M1 **laptops**) wasn't working correctly. The tests were hanging indefinitely, and it was needed to close the electron application manually for each one to make it pass.

# Approach chose to fix this

When running each test **individually** it was detected that one of them `(actionValues.test.ts`) was **always** working.

That test is one of the few tests using the statement `await electronService.clickStopRecording(); `

It turns out that the absence of that line on the other tests was the main reason behind the problem. That's why the `await electronService.terminate();` inside the `afterEach` wasn't responding. Although not confirmed, I would dare to say that we were facing a memory issue.

I'm sharing two recordings. 

**Before the fix**

https://user-images.githubusercontent.com/15065076/210569259-fd5ebef7-db6e-404b-928f-9bcf78f104f3.mov

**After the fix** 🎉 

https://user-images.githubusercontent.com/15065076/210567959-3740b6f2-ce6d-4324-9182-536d3fc266ed.mov

### DIFF
--- a/e2e/tests/dragAndDrop.test.ts
+++ b/e2e/tests/dragAndDrop.test.ts
@@ -69,6 +69,7 @@ describe('Drag and Drop', () => {
     await electronService.recordClick('text=Hello Elastic Synthetics Recorder');
 
     await (await electronWindow.$('id=insert-divider-0-1')).click();
+    await electronService.clickStopRecording();
     await (await electronWindow.$('id=step-1')).hover();
     await electronWindow.mouse.down();
     await electronWindow.mouse.move(100, 100, { steps: 5 });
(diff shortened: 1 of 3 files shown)


kyungeunni:
@devcorpio In fact I've tried that in the past but here's what happens on my local machine. I've pulled your branch and ran tests:
  image

devcorpio:
strange. I recommend a pair session to tackle this properly

justinkambic:
Skipping Post-FF testing as this was not merged.

kyungeunni:
This was closed as it was resolved by https://github.com/elastic/synthetics-recorder/pull/346

PR #346, opened 2023-01-05
fix: close recording target browsers before terminating electron when testing

kyungeunni:
## Summary
When running e2e tests, make sure the recording target browser is closed before closing electron instance so that each test can be finished. Currently, the second instance of electron starts and hangs forever when the browser is not closed.
(It seems to be an issue only to Mac)

## Implementation details
Close all the contexts and browser before terminating electron. But the graceful termination of browser doesn't guarantee the complete close of the browser so adding setTimeout hack was needed

### DIFF
--- a/e2e/services/browser.ts
+++ b/e2e/services/browser.ts
@@ -96,6 +96,12 @@ export class TestBrowserService {
 
   static async closeRemoteBrowser(connectParams?: ConnectRetryParams) {
     const browser = await TestBrowserService.getRemoteBrowser(connectParams);
+    const ctxs = await browser.contexts();
+    for (const ctx of ctxs) {
+      await ctx.close();
+    }
     await browser.close();
+    // with this timeout, it will make sure the browser to be closed before terminating electron (it's an issue to mac).
+    await new Promise(r => setTimeout(r, 50));
   }
 }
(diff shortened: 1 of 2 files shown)


devcorpio (approved):
LGTM! 🎉

e2e/services/browser.ts
```diff
@@ -96,6 +96,12 @@ export class TestBrowserService {
 
   static async closeRemoteBrowser(connectParams?: ConnectRetryParams) {
     const browser = await TestBrowserService.getRemoteBrowser(connectParams);
+    const ctxs = await browser.contexts();
+    for (const ctx of ctxs) {
+      await ctx.close();
+    }
     await browser.close();
```
vigneshshanmugam, line 103: closing browser, closes the underlying context. Am i missing something?

kyungeunni: @devcorpio and I tried running tests with closing browser only, and closing contexts first then closing the browser(we ran 5 consecutive test runs for both cases). Closing contexts first were more reliable than the other.

vigneshshanmugam: That seems like a bug to me in PW. I have seen in their code that they dont close the context as well.

justinkambic:
Post-FF Testing LGTM

Issue #264, opened 2022-06-01
Add more unit tests


justinkambic:
Thus far we have been approaching unit testing with the mindset of "don't write tests for anything we think might change". This is primarily related to UI components, but there are a number of untested backend functions and other areas of the project that we should cover before going GA.

It's possible this ticket will eventually become a meta issue for tracking different segments of the codebase. We might also want to improve our automation to provide easy access to metrics on unit test coverage, so we can zero in on areas that are completely untested.

## Todo
- [ ] Add tests to `onTest` electron code https://github.com/elastic/synthetics-recorder/issues/235
- ...

kyungeunni:
Thanks to @justinkambic, Frontend code has very good test coverage, but we don’t have the test coverage we’d like to have on the main process part yet. However, e2e tests complement the lack of unit tests on this part so we will close this for now, and come back to it.

PR #358, opened 2023-01-26
Add unit tests for `<TestResult />`

justinkambic:
## Summary

**NOTE TO REVIEWERS:** This should not be reviewed until #353 is merged.

Related to #264. Adds test for `<TestResult />`.

## Implementation details

Adds unit tests.

If the test looks like it makes sense and it is passing CI no further testing needed.

### DIFF
--- a/src/components/TestResult/TestResult.test.tsx
+++ b/src/components/TestResult/TestResult.test.tsx
@@ -1,0 +1,208 @@
+/*
+MIT License
+
+Copyright (c) 2021-present, Elastic NV
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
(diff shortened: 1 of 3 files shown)


src/components/TestResult/TestResult.test.tsx
```diff
@@ -0,0 +1,208 @@
...
+*/
+
+import { fireEvent, waitFor } from '@testing-library/react';
+import React from 'react';
+import { createStep } from '../../../common/helper/test/createAction';
+import { render } from '../../helpers/test';
+import { TestResult } from './TestResult';
+
+describe('<TestResult />', () => {
+  it('does not render the flyout if there is no result data', () => {
```
kyungeunni, line 32: I'm wondering if it makes sense to improve this by showing _Waiting for tests result.._ or something similar rather than not showing it at all. But this is just an idea and out of the scope of this PR

kyungeunni (approved):

src/components/TestResult/TestResult.test.tsx
justinkambic: Hm, if testing is in progress and we're awaiting the result, that could be a better UX. I'll add a ticket to address that post v1.

justinkambic: https://github.com/elastic/synthetics-recorder/issues/365

Issue #265, opened 2022-06-06
Provide license info with the Script Recorder


paulb-elastic:
As part of making the Script Recorder GA, we need to declare the licenses of the various components we're using (e.g. in an _About Box_)

kyungeunni:
Had a chat with the team. We will create a txt file declaring the licenses (similar to how [beats](https://github.com/elastic/beats/blob/main/NOTICE.txt) or [synthetics](https://github.com/elastic/synthetics/blob/main/NOTICE.txt) do) and bundle it together, and show it in the _About_ box, either as textarea-like element or acknowledgments button that opens up a window with the license contents in it. 
example of elastic SHIELD: 
<img width="649" alt="image" src="https://user-images.githubusercontent.com/16660866/185438178-66c4fc0c-5fa2-47f0-96b1-37ca29bb46c3.png">

justinkambic:
Post-FF Testing LGTM

PR #361, opened 2023-01-30
feat: generate NOTICE.txt

kyungeunni:
part of #265, [#346](https://github.com/elastic/open-source/issues/346)

## Summary
Generate NOTICE.txt listing dependencies used by Synthetics Recorder. Running the script in postinstall hook so whenever new deps are added, we will update the NOTICE.txt too.

## Implementation details
Borrow [script](https://github.com/elastic/apm-agent-rum-js/blob/main/dev-utils/dep-info.js) from 
[apm-agent-rum-js](https://github.com/elastic/apm-agent-rum-js)
- [x] List npm dependencies that will be bundled in the build
- [x] Include chromium license which is not part of npm dependency

- Should see `NOTICE.txt`in the repo
- ~Should see extended `NOTICE.txt` that includes platform-specific binaries included in each artifact's resources dir~ : Figured we can simplify this by just including chromium license in NOTICE.txt


dev-tools/dep-info.js
```diff
@@ -0,0 +1,123 @@
...
+    );
+  }
+  const license = Buffer.from(bodyStr, 'base64').toString('utf-8');
+  return { name: 'chromium', license };
+}
+
+async function generateNotice() {
+  const { name: pkgName, dependencies } = getMeta();
+  const chromiumInfo = await getChromiumInfo();
+  const depInfo = [chromiumInfo, ...generateDependencyInfo(Object.keys(dependencies))];
```
kyungeunni, line 99: At first, I wanted to generate the NOTICE.txt file that includes only npm production dependencies, then generate the extended NOTICE.txt including Chromium license when building the app. But I figured we might simplify it by keeping only one NOTICE.txt in the repo and copy it over when we release the app.

WDYT @justinkambic @vigneshshanmugam ?

justinkambic: I'll defer to @vigneshshanmugam as to whether we should have the chromium license included, but I am fine with bundling that all together when we release the app. With the size of the binary we're going to ship, including a full browser, some extra lines of license copy are not going to move the needle very much on dist size. That's my impression, at least.

dev-tools/dep-info.js
```diff
@@ -0,0 +1,123 @@
...
+const ROOT_NODE_MODULES = join(ROOT_DIR, 'node_modules');
+const CHROMIUM_LICENSE_URL =
+  'https://chromium.googlesource.com/chromium/src/+/refs/heads/main/LICENSE?format=TEXT';
+
+function safeReadDirSync(directory) {
+  let files = [];
+  try {
+    files = readdirSync(directory);
+  } catch (_) {
+    // do nothing
```
justinkambic, line 39: Could we assert the error is of the type(s) we want to drop on the floor, and otherwise throw the error?

```typescript
// something like this, although I didn't actually test this
if (error.text !== 'This is the error I want to ignore') throw error;
```

NOTICE.txt
```diff
@@ -0,0 +1,442 @@
+
+synthetics-recorder
+Copyright (c) 2017-present, Elasticsearch BV
```
justinkambic, line 3: Re-running the script updates this to `2021-present`, so we might want to overwrite this with the latest output of the script.

dev-tools/dep-info.js
kyungeunni: Oh let me clarify my intention. My initial approach was:
1. We keep `NOTICE.txt` listing npm production deps in the repo
2. When we release the app, add the chromium license to the existing `NOTICE.txt`. Now we have created (temp)NOTICE.txt
3. Copy over the (temp)NOTICE.txt as a build resource

and my question was that instead of taking step 2, we add the chromium license info in step 1 and remove step 2 completely(That's what I did in this PR too).

But now that I think about it, my initial approach doesn't really make sense as we don't need to make a distinction between node modules and chromium.

vigneshshanmugam: ++ Just keeping a single Notice.txt would be better IMO.

dev-tools/dep-info.js
vigneshshanmugam: We could just log though, not worth in throwing error when the dep doesnt contain license info in License file or Notice file.

vigneshshanmugam (approved):
LGTM

dev-tools/dep-info.js
kyungeunni: ++ I think we want it to silently fail here.

NOTICE.txt
kyungeunni: Thanks, updated!

kyungeunni:
@justinkambic I have another PR coming up that is built up on this PR. I will merge it for now as I see most of the comments are resolved. Let me know if you have any other concerns!!

PR #371, opened 2023-02-01
Delete unused code

justinkambic:
While writing unit tests I noticed this code which does not appear to be referenced elsewhere in the app.

I'm confused why the unused exports check didn't recognize this, as it just flagged me on another branch.

In any case, I think it's safe to destroy this code.


kyungeunni (approved):
Oh, thanks for spotting this. It is indeed weird that it hasn't been detected by our tooling.

Issue #235, opened 2022-04-29
[Other] Add tests to `onTest` electron code


justinkambic:
### Task summary

We recently [fixed](https://github.com/elastic/synthetics-recorder/pull/233) a bug that caused the error messages from Playwright not to show in the recorder. This happened because we merged a change that removed an optional field from the result object.

As a follow-up to this fix we should focus on refactoring the `execution.ts` file to be more testable and add tests to ensure the output remains consistent when subjected to changes.

### Linked resources

PR that inspired the issue: https://github.com/elastic/synthetics-recorder/pull/233

justinkambic:
_stale orphaned tickets_
_they wait perpetually_
_rusting derelicts_

PR #376, opened 2023-02-07
[chore] Apply context isolation and rearrange `execution.ts`

kyungeunni:
base work for 
- #264 
- #235

## Summary

Currently, all the functions running in the main process were mangled together in the `electron/execute.ts` file. This makes it quite difficult to understand, hard to maintain, and a blocker for adding tests. 

This PR tries to extract each function to its own file so that we could avoid the difficulties mentioned above, and make it possible to add unit tests.

Another important thing is that this PR introduces [context isolation](https://www.electronjs.org/docs/latest/tutorial/context-isolation) as per the [security guideline](https://www.electronjs.org/docs/latest/tutorial/security#3-enable-context-isolation) provided by Electron.

## Implementation details
- Turn off node integrations and enable context isolation
- functions in `execution.ts` is now divided into files under `api/` directory
- the whole ipc is no longer exposed in CommunicationContext, now limited set of `window.electronAPI` is exposed

run test, run dev

**Note to reviewers**: 
 I'm afraid this PR is quite big to go through, I wish I could break it down into smaller chunks of PRs but it was inevitable to do it all at once as everything was connected. I recommend seeing how the communication between renderer process <-> main process has been changed by exposing selected APIs in `electron/preload.ts` to align with the [best practices](https://www.electronjs.org/docs/latest/tutorial/context-isolation#security-considerations).
Also, there are new components like `browserManger` or `syntheticsManager`. Those are introduced as it requires a somewhat shared state between functions as well as to make it easier for us to mock the dependency when testing. That being said, it might be an intermediate change that is subject to change in upcoming tests PRs.


electron/api/runJourney.ts
```diff
@@ -0,0 +1,214 @@
...
+  RunJourneyOptions,
+  StepEndEvent,
+  StepStatus,
+  TestEvent,
+} from '../../common/types';
+
+import { JOURNEY_DIR, PLAYWRIGHT_BROWSERS_PATH } from '../config';
+import { SyntheticsManager } from '../syntheticsManager';
+
+export async function runJourney(
```
kyungeunni, line 41: This function needs some more attention. But I prefer to defer it as I need to add unit tests for this and that'd be better time doing that.

package.json
```diff
@@ -85,7 +85,6 @@
     "@emotion/cache": "^11.9.3",
     "@emotion/react": "^11.9.3",
     "dotenv": "^16.0.3",
-    "electron-better-ipc": "^2.0.1",
```
kyungeunni, line 88: Don't need it anymore. The same feature is now supported in electron

electron/preload.ts
```diff
@@ -0,0 +1,76 @@
...
+    ipcRenderer.on('test-event', callback);
+    await ipcRenderer.invoke('run-journey', params);
+  },
+
+  removeOnTestListener: () => {
+    ipcRenderer.removeAllListeners('test-event');
+  },
+};
+
+contextBridge.exposeInMainWorld('electronAPI', electronAPI);
```
justinkambic, line 76: This is where we get `window.electronAPI` from on the renderer process then?

src/helpers/test/mockApi.ts
```diff
@@ -22,33 +22,20 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 */
 
-import type { RendererProcessIpc } from 'electron-better-ipc';
+import { IElectronAPI } from '../../../common/types';
 
-export function getMockIpc(overrides?: Partial<RendererProcessIpc>): RendererProcessIpc {
+export function getMockElectronApi(overrides?: Partial<IElectronAPI>): IElectronAPI {
```
justinkambic, line 27: 👌

src/hooks/useRecordingContext.test.ts
```diff
@@ -71,20 +69,20 @@ describe('useRecordingContext', () => {
...
+      useRecordingContext(electronApi, '', 0, setResult, setSteps)
     );
     const recordingContext = renderHookResponse.result.current;
 
     await act(async () => recordingContext.toggleRecording());
 
     renderHookResponse.rerender();
 
-    expect(callMainMock).toHaveBeenCalledWith('record-journey', { url: '' });
+    expect(recordJourney).toHaveBeenCalledWith('');
```
justinkambic, line 80: It's outside the scope of this patch but could you do me a favor and declare a `url` at the start of this test, supply it to the hook, and verify that `recordJourney` was called with that URL? I think now that the args the mock receives are just a string it has made the test's intentions less obvious, since the assertion has become just an empty string.

```typescript
const url = 'https://elastic.co';

// ...

useRecordingContext(electronApi, url, setResult, setSteps);

// ...

expect(recordJourney).toHaveBeenCalledWith(url);
```

src/hooks/useSyntheticsTest.ts
```diff
@@ -56,42 +56,45 @@ export function useSyntheticsTest(steps: RecorderSteps): ITestContext {
...
+      const code = await getCodeFromActions(electronAPI, steps, 'inline');
       if (!isTestInProgress) {
         // destroy stale state
         dispatch({ data: undefined, event: 'override' });
         const onTestEvent = (_event: IpcRendererEvent, data: TestEvent) => {
           dispatch(data);
         };
 
-        ipc.on('test-event', onTestEvent);
+        // electronAPI.on('test-event', onTestEvent);
```
justinkambic, line 67:
```suggestion
```

We want this removed now? I am assuming it's ok since the tests are passing, will confirm when I smoke test.

common/types.d.ts
```diff
@@ -115,3 +115,20 @@ export type GenerateCodeOptions = {
...
   isProject: boolean;
 };
+export interface IElectronAPI {
+  exportScript: (string) => Promise<boolean>;
+  recordJourney: (url: string) => Promise<void>;
+  stopRecording: () => void;
+  pauseRecording: () => Promise<void>;
+  resumeRecording: () => Promise<void>;
+  onActionGenerated: (
+    callback: (event: IpcRendererEvent, actions: ActionContext[]) => void
```
justinkambic, line 125:
```suggestion
    callback: ActionGeneratedCallback
```

I mention this further down in the review, but I think it might be nice to do something like this:

```typescript
type ActionGeneratedCallback = (event: IpcRendererEvent, actions: ActionContext[]) => void;
```

Then we can reference this type in the implementation of `onActionGenerated`, and where we set it up in `preload.ts`. It gives a little shorthand for defining the argument type.

common/types.d.ts
```diff
@@ -115,3 +115,20 @@ export type GenerateCodeOptions = {
...
+  ) => () => void;
+  generateCode: (params: GenerateCodeOptions) => Promise<string>;
+  openExternalLink: (url: string) => Promise<void>;
+  runTest: (
+    params: RunJourneyOptions,
+    callback: (event: IpcRendererEvent, data: TestEvent) => void
```
justinkambic, line 131: Same as the above comment regarding callback typing.

justinkambic (commented):
Had a few slight improvements/asks and some nit comments.

Really fantastic work on this!

electron/preload.ts
```diff
@@ -0,0 +1,76 @@
...
+THE SOFTWARE.
+*/
+import type {
+  ActionContext,
+  GenerateCodeOptions,
+  IElectronAPI,
+  RunJourneyOptions,
+  TestEvent,
+} from '../common/types';
+import { contextBridge, ipcRenderer, IpcRendererEvent } from 'electron';
```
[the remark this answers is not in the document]
kyungeunni: Same here! I was reading through the Electron doc and realized that we are not following the recommended way. I'm glad we had a chance to address it this time! 😊

electron/preload.ts
kyungeunni: yes it is!

src/hooks/useSyntheticsTest.ts
kyungeunni: Oops, yes this can be removed here and the [listener](https://github.com/elastic/synthetics-recorder/pull/376/files#diff-17fc44a47be3f3309d4ce97948f68135a3ca438711a75d645a0e3710aa92098bR67) will be callled within `electron.runTest`. Good spot!!

common/types.d.ts
kyungeunni: Thanks for this comment! While applying your suggestion, I've observed an interesting DX case I'd like your opinion on. I usually prefer the way suggested, but this time it seems like it is quite useful to have ad-hoc type definition when implementing this function elsewhere;  If I define the type for the callback and hover when using the function, it will make me go over to check the details of `ActionGeneratedCallback`type definition:

image

But when it's defined inline, I can check the type definition of the callback straight away without hopping the definition chains:

image

Also, I noticed that I didn't need to define the argument type again in `preload.ts` so we don't need to reference this type again in there.

WDYT @justinkambic ?!

src/hooks/useRecordingContext.test.ts
kyungeunni: Yup I've updated as you suggested and the test seems to work!

common/types.d.ts
justinkambic: Sure, that is fine with me as well. Given that this signature is only expected a handful of times in this case I think it's an ok tradeoff.

If, for example, we were making a library where we were expecting to reference this signature an arbitrary amount of times, I'd say we should probably define a custom type. It can have added DX burden if you need to drill into the types, but I think it's generally cleaner when reading the code. I'm curious what happens if you hover on `callback` in this case, does it only show `ActionGeneratedCallback` as well?

src/hooks/useRecordingContext.test.ts
justinkambic: Thank you, I think it makes the test more readable.

justinkambic (approved):
LGTM!

common/types.d.ts
kyungeunni: When you hover on `callback` (I renamed it to `listener` here), it shows type definition that I specified for the callback, but as soon as I put the wrong types and try to use it in `electronAPI.addActionGeneratedListner`, it complains:
https://gyazo.com/5759b1b9c960bcde61f4c39ab8e99669

PR #377, opened 2023-02-07
Add unit tests for `ExportFlyout`

justinkambic:
## Summary

Related to #264. Adds test for `ExportFlyout`.

## Implementation details

Adds unit tests.

If the test looks like it makes sense and it is passing CI no further testing needed.

### DIFF
--- a/src/components/ExportScriptFlyout/Body.test.tsx
+++ b/src/components/ExportScriptFlyout/Body.test.tsx
@@ -1,0 +1,68 @@
+/*
+MIT License
+
+Copyright (c) 2021-present, Elastic NV
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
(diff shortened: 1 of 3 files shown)


kyungeunni (approved):
LGTM! 

I've noticed that there are some console errors in the test result, I'm not sure it's introduced with this PR but just letting you know..!!
image

justinkambic:
> I've noticed that there are some console errors in the test result

@kyungeunni thanks for pointing that out. I think this is mostly happening because EUI components tend to do a pass-through of unknown props to their child elements, and sometimes this includes DOM elements like `div`.

For example, in the tests for `SeparatorActions`, we declare a prop `isGrabbing` on a custom-styled `EuiFlexGroup`. Then, `EuiFlexGroup` [passes that prop](https://github.com/elastic/eui/blob/16b0b299a1810b5a35f79f46d22da7de38101dd0/src/components/flex/flex_group.tsx#L118) as part of `rest` to the underlying DOM node.

I think this is not a huge deal at the moment, but I'll create a ticket to take care of these. The easiest solution is to create a custom React component that wraps the `EuiFlexGroup`, but does not pass the custom props down to it. This will get rid of the warnings while having essentially no impact on the UX. I'll try to catalog all the places where this is happening and collate them in the ticket.

kyungeunni:
@justinkambic 

I missed this but I'm also seeing this in the console:
image

should we make these test that prints console.error fail so we have visibility on these?

justinkambic:
@kyungeunni I noticed this as well. If you like we can temporarily skip these tests and update the config to fail on console.error, that is fine with me.

PR #391, opened 2023-03-29
Upgrade EUI

justinkambic:
## Summary

Upgrades EUI version to v76.

## Implementation details

There should be minute or no GUI changes, and the app should pass all CI tests.

image

For this particular screen, the header remains un-dimmed, whereas in previous versions of EUI it was dimmed. I've conferred with the EUI team about this and they have told me it's working as designed and although it's different from how it looked before it isn't violating any of their design principles.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
   },
   "dependencies": {
     "@elastic/datemath": "^5.0.3",
-    "@elastic/eui": "^58.1.1",
+    "@elastic/eui": "^76.4.0",
     "@elastic/synthetics": "=1.0.0-beta.38",
     "@emotion/cache": "^11.9.3",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -80,7 +80,7 @@
   },
   "dependencies": {
     "@elastic/datemath": "^5.0.3",
-    "@elastic/eui": "^58.1.1",
+    "@elastic/eui": "^76.4.0",
```
kyungeunni, line 83: 🎉

kyungeunni (approved):
Tested locally, all UI components seems to be intact. Thanks for confirming with EUI team regarding undimmed header, it does seem a bit odd but I'm ok with it as that's our design decision 👍

PR #403, opened 2023-04-19
[UI] Make page title sticky and remove box shadow

### LINKED ISSUE #396: [Bug] The header is hidden when scrolls down
### Bug summary

Seems like it's introduced by #391 

_Elastic Synthetics Recorder_ header should be fixed at the top and shouldn't be scrolled with the main contents:

[Image from Gyazo](https://gyazo.com/7bc0c4a481043754618fcd95dda1b6c2)

### Recorder Version

main branch

### Steps to reproduce

1. Create a script that has many actions that would make the scroll bar appear.
2. Scroll down and you can see the header is also scrolled together.

### Expected behavior

The header should be fixed at the top all the time

### Additional information

_No response_

justinkambic:
## Summary

Resolves #396 .

Makes the recorder's page header static at the top of the viewport during scrolling, and removes its box shadow.

## Implementation details

When we upgraded EUI we encountered some less-than-desired behavior (shown in the linked issue) where an expanded flyout leaves the title section of the page undimmed while the flyout is expanded. The true intention of this functionality is that the flyout will appear below a static heading UI, whereas ours fills in all the way to the top of the viewport.

To resolve this unwanted behavior, I've modified the page heading to be sticky, per the design standards shown in both the EUI docs and in multiple parts of Kibana. Now, when the user scrolls, the page heading will remain at the top of the viewport, and will remain accented when the flyout opens.

A side effect of this change is the removal of the box shadow; because it is highlighted while the flyout is open, it makes the title look weird. Rendering the shadow conditionally is also less-than-ideal because having it snap away when the flyout opens is also janky.

You can compare this change by running the recorder in dev mode and switching between this branch and main.

### DIFF
--- a/src/components/Header/HeaderControls.tsx
+++ b/src/components/Header/HeaderControls.tsx
@@ -38,8 +38,8 @@ import { UrlField } from './UrlField';
 const Header = styled(EuiFlexGroup)`
   background-color: ${props => props.theme.colors.emptyShade};
   border-bottom: ${props => props.theme.border.thin};
+  padding: 65px 16px 16px 16px;
   margin: 0px;
-  padding: 16px;
 `;
 
 const TestButtonDivider = styled(EuiFlexItem)`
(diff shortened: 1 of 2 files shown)


kyungeunni:
Tested locally, and the header is now fixed at the top 👍 
For some reason, I do not see the `Journey Code` header in Export flyout though
image

kyungeunni (approved):
LGTM!
image

Issue #393, opened 2023-03-30
Update the Synthetics/Playwright versions in the Script Recorder


paulb-elastic:
For the next (1.0) release, we need to update Synthetics and Playwright to a more recent version, more aligned with the 8.8 release

- [ ] Update Playwright and Synthetics ([reference](https://github.com/elastic/synthetics-recorder#managing-playwright-dependency))
- [x] [Review the code generation](https://github.com/elastic/synthetics-recorder/issues/393#issuecomment-1495288960) and bring it up-to-date with the updated version of Playwright

kyungeunni:
I'm adding one more task to review the [code generation part](https://github.com/microsoft/playwright/blob/release-1.31/packages/playwright-core/src/server/recorder/javascript.ts) from the Playwright when updating the new version, as this is copied manually from Playwright's code base and residing in our [codebase](https://github.com/elastic/synthetics-recorder/blob/main/electron/syntheticsGenerator.ts#L151). Seems like our generator is outdated.

vigneshshanmugam:
@kyungeunni Can you do it on the [Synthetics](https://github.com/elastic/synthetics/blob/f5dc2af66f8e0859ba3d89540235682bb930e386/src/formatter/javascript.ts#L75) given we can go back and use it instead of having it here. We copied it here as we wanted to use a custom steps stuff. Now that both are pretty much the same, we can go ahead and have it on single place instead. Thanks.

kyungeunni:
@justinkambic Do you think we can tackle the above point @vigneshshanmugam mentioned as part of updating Playwright version update? I think it's also worth writing it down in the README.md about the codegen as part so we don't forget.

justinkambic:
I think it would be good to follow-up [the existing beta-43 upgrade](https://github.com/elastic/synthetics-recorder/pull/399) with this migration, as that upgrade isn't going smoothly and I'd prefer to limit scope creep.

justinkambic:
I'm re-opening this. We closed the ticket after a code-gen update, but I am going to re-open this as we can now track the migration for the recorder to Synthetics v1 after it is released.

We also might want to keep this open until https://github.com/elastic/synthetics-recorder/issues/407 is resolved as well.

justinkambic:
[...]

PR #406, opened 2023-04-28
chore: bring codegen up-to-date

kyungeunni:
## Summary
- Bring codegen up-to-date as [Playwright](https://github.com/elastic/playwright/blob/1.31.1-recorder/packages/playwright-core/src/server/recorder/javascript.ts#L28)
-  Add updating codegen section in README.md 

Resolves https://github.com/elastic/synthetics-recorder/issues/393.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,9 @@ git push  --set-upstream elastic 1.20.1-recorder
 
 For now, we are doing it all manually, However, we should consider automating the process.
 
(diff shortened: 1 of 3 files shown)


justinkambic (approved):
Some minor grammar nit on the README update and I deleted a function that was no longer used. The generated code seems in keeping with what we were doing already, and appears to work.

LGTM

PR #408, opened 2023-05-03
[Chore] Bump EUI and styled-components

justinkambic:
## Summary

Updates `EUI` and `styled-components` to use more recent versions.

## Implementation details

There should be no noticeable delta in terms of features or use.

Run the recorder and ensure that while smoke testing nothing seems off. If e2e tests are passing and everything looks visually ok, this is safe to merge.


vigneshshanmugam (approved):
LGTM

PR #422, opened 2023-05-18
[Chore] Bump deps post-v1

justinkambic:
## Summary

Bumps dependencies for most packages. We didn't do this before v1 so as to avoid any unforeseen instability.

## Implementation details

Biggest question mark is updating `electron`, there may be some breaking change impacts as we jumped 20-24.

Make sure the recorder is working. We can do a release build to try and make sure as well.


.gitignore
```diff
@@ -13,3 +13,5 @@ eslint-junit.xml
 *-junit.xml
 .DS_Store
 tsconfig.tsbuildinfo
+dump.rdb
```
vigneshshanmugam, line 16: Ah did you try running `redis-cli` from the recorder repo path? Not sure if we need this here anyway.

vigneshshanmugam (approved):
LGTM

package.json
```diff
@@ -93,7 +93,7 @@
     "playwright": "https://gitpkg.now.sh/elastic/playwright/packages/playwright-core?1.32.3-recorder",
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
-    "styled-components": "^5.3.7",
+    "styled-components": "^6.0.0-rc.1",
```
vigneshshanmugam, line 96: is this worth doing right now given its a release candidate? Also why not just go to rc.3 if we want to proceed with the RC.

justinkambic: I think I upgraded for `npm audit` reasons. Agreed though, will downgrade 👍

.gitignore
justinkambic: It's possible but typically not. I'll remove it anyway as I've never needed it before, probably something weird I did.

package.json
justinkambic: Fixed c7782d57f90684101fb7fa16e4f98b510fad26f6

.gitignore
justinkambic: Fixed c7782d57f90684101fb7fa16e4f98b510fad26f6

PR #437, opened 2023-06-29
[Fix] Make CI builds pass again

justinkambic:
## Summary

Resolves https://github.com/elastic/observability-robots/issues/1750.

We have been struggling with broken release CI builds.

This patch introduces a new step to the `download-browsers` script that will ensure all of the contents of the browser packages are readable and writable by the Jenkins user that is responsible for building the Recorder.

### Implementation details

In a newer version of Chrome (required by an updated Playwright for this version) there is a file that does not unpack as writable, which causes our code sign/notarization piece to fail during the release pipeline. These types of errors tend to only be found late in the release process because we don't perform these steps as part of our typical builds.

### How to validate this change

You can push a tag and view the Jenkins pipeline to see if the release build succeeds.

### NOTE

This does also downgrade the `electron` and `electron-builder` versions to what they were in the previous release. This is because their env expectations have changed and, while it would be nice to upgrade, there was no pressing reason to upgrade in terms of features, bugs, or security, so rather than also troubleshoot those issues at the same time I am planning to delay that pain for a future release.


scripts/download-chromium.js
```diff
@@ -98,4 +99,52 @@ exports.downloadForPlatform = async function downloadForPlatform(electron_platfo
     getChromeVersion()
   );
   await download(platform, arch, revision, directory);
+  setPermissions(directory);
```
emilioalvap, line 102:
```suggestion
  await setPermissions(directory);
```

scripts/download-chromium.js
```diff
@@ -98,4 +99,52 @@ exports.downloadForPlatform = async function downloadForPlatform(electron_platfo
...
 };
+
+async function setPermissions(directory) {
```
vigneshshanmugam, line 105: the error handling on this function seems really off, I would suggest to bubble up the error and handle with a single try/catch logic.

justinkambic: It's what happens when you spend hours figuring out why a CI build is failing, and then write a quick hack to try and fix it 😄

vigneshshanmugam (approved):

PR #446, opened 2023-07-12
Bump EUI to 84.0.0

justinkambic:
## Summary

Bumps the EUI version to 84.0.0.

## Implementation details

UI should look identical to what we shipped in v1.1.0.

Run the recorder, preferably alongside the previous release version, make sure the EUI elements look the same.


justinkambic:
We just bumped to v89

PR #472, opened 2023-10-27
V1.3.0 upgrade prep

justinkambic:
Bumps dependencies in preparation for bumping Synthetics and releasing v1.3.0 of the recorder.


vigneshshanmugam (approved):
LGTM

We bumped Synthetics to `18.18.2` Node version due to a CVE. You might need to prep for that as well. FYI

PR #480, opened 2023-11-08
Bump synthetics 1.5.3

justinkambic:
## Summary

**NOTE:** During review we discovered that the mismatched semver between `playwright` and `playwright-core` was not intentional, so we are going to wait for an updated version of Synthetics to be released, and then we will update this PR to reference that and unify the semver for the Playwright deps.

### What's going on here

Previously we have always referenced `playwright-core` as `playwright` in the recorder, because we only leverage features from core. The later versions of Synthetics now reference the `esmLoader` file from the `playwright` package, and the old build can no longer work. So, I have introduced both subpackages to the Recorder's dependency list here.

~One other thing to note is the Elastic PW fork we use now has versions for both `1.38.0` and `1.38.1`. This is because Synthetics references the `*.1` patch for core.~

**NOTE:** the inconsistency with Playwright package versioning has been resolved by an upstream release, but I am leaving the instructions as-is barring feedback to convince me otherwise, as this could be a contingency in the future in which those docs would be helpful.

A nice side effect is that, previously we were hacking Synthetics code to reference the core package for a certain case. We no longer need to do that now that we reference both packages and they correlate to the correct place that Synthetics expects to use.

Lastly, I have added significant supplemental info to the project's `README.md` documenting this new process for `playwright` and `playwright-core`.

## Implementation details

Shouldn't be any dependency issues, we are just going to update deps and not introduce new functionality.

If CI is passing all should be good. It would be good to pull the changes and do an `npm ci` to make sure you can record and re-run journeys and export generated code.


vigneshshanmugam (approved):
LGTM. Love the detailed steps on the Readme. Great Improvement 🎉

PR #481, opened 2023-11-10
feat: windows sign endpoint to http://timestamp.digicert.com

v1v:
## Summary

http://timestamp.acs.microsoft.com was down so let's use http://timestamp.digicert.com. That endpoint has been used in the Unified Release process for quite sometime and it's stable

Interestingly the metadata file shows the other azure endpoint though things worked fine

image

In the CI, see this [build](https://internal-ci.elastic.co/job/elastic+synthetics-recorder+release/view/tags/job/v1.3.0-rc8/)


v1v:
PR:
- https://github.com/elastic/synthetics-recorder/pull/481

vigneshshanmugam (approved):
Great find 🎉 LGTM

justinkambic:
Tried this as well on https://github.com/elastic/synthetics-recorder/pull/480, it turns out that the microsoft server we've always used was unavailable for quite a long time, but the service is working again. Closing this as we no longer need it.

Thanks again @v1v for your super fast efforts to help resolve this ❤️

vigneshshanmugam:
@justinkambic Why dont we need this anymore? Could you help me understand the reasoning?

justinkambic:
> Why dont we need this anymore? Could you help me understand the reasoning behind it. Thanks!

@vigneshshanmugam sorry just saw this. When I said "Closing this as we no longer need it.", I meant we don't need any revision to the _existing_ codesign pipeline, not that we don't need it at all. So if that was your concern then fear not, we are still performing Windows signing as part of our release pipeline.

We experienced some failures because the normal timestamp server the Windows sign step used was unavailable, but after a  few days it was back up again and we could use it once more. So we didn't need to switch to the digicert one. You can see that change reflected [here](https://github.com/elastic/synthetics-recorder/pull/481/files#diff-9e17a05eae036985743857ae5658a4cbe09ca8a65352d464eed7b147e3aa84e9R1) in the PR diff, where the only update is to the timestamp server URL. Since the old server was available again and the code sign worked as it always has, we can close this PR.

PR #482, opened 2023-11-13
[Chore] Update release docs

justinkambic:
## Summary

Updates to the Release and Downloads docs in anticipation of the upcoming v1.3.0 release.

## Implementation details

Provides some testing guidance for the basic quality gate process we have in place before we publish a release.

Also updates the Downloads page to work with the new release.

This should not be merged until the release is published.

## How to validate this change

Check for grammatical errors and make sure the instructions are clear.

PR #483, opened 2023-11-14
Merge v1.3.0 to main

justinkambic:
I am no longer capable of pushing directly to upstream `main`, so making this PR to merge the change that contains the update. I think this is ok, since the release tag is what we should go off of for the release.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "synthetics-recorder",
-  "version": "1.2.0",
+  "version": "1.3.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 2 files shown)


vigneshshanmugam (approved):
LGTM. I saw you did the same on the release docs PR https://github.com/elastic/synthetics-recorder/pull/482 Is this intended?

justinkambic:
>  I saw you did the same on the release docs PR

This commit is present on that one too. I made this change initially on my local `main`, because in the past we do it that way and push to upstream, but since that is now blocked, when I made the docs PR it spawned off of a `main` that was ahead of the remote. So, after I merge this, I will merge it into my local `main` and then rebase the docs PR off of that, the diff will update.

PR #484, opened 2023-11-16
chore(github-action): create test check status for docs only

v1v:
## Summary

```cucumber
Given the Test workflow (.github/workflows/test.yml) with paths-ignore
When the GitHub check status called `test` is enabled as one of the rules in the Branch protection
Then it's required to synthetically set the GitHub check status called `test` for those changes that are in the `paths-ignore'
```

## Implementation details

A new GitHub workflow that overrides the name of the GitHub check and listening for the files that have been [ignored in the test GitHub](https://github.com/elastic/synthetics-recorder/blob/a6f4b6cb915d94a74a06247e379c8a141f32af34/.github/workflows/test.yml#L6-L10) workflow.

Once it's merged

## Issues

Unblocks https://github.com/elastic/synthetics-recorder/pull/482

### DIFF
--- a/.github/workflows/test-docs.yml
+++ b/.github/workflows/test-docs.yml
@@ -20,6 +20,6 @@ permissions:
 
 jobs:
   test:
-    runs-on: macos-latest
+    runs-on: ubuntu-latest
     steps:
       - run: 'echo "No build required"'
--- a/.github/workflows/test-docs.yml
+++ b/.github/workflows/test-docs.yml
@@ -1,0 +1,25 @@
+---
+# This workflow sets the Test / test status check to success in case it's a docs only PR and test.yml is not triggered
+# https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks
+name: Test # The name must be the same as in test.yml
+
+on:
+  push:
+    branches:
+      - main
+    paths-ignore: # This expression needs to match the paths ignored on test.yml.
+      - '**'
+      - '!**/*.md'
+  pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/test-docs.yml
```diff
@@ -0,0 +1,25 @@
...
+  push:
+    branches:
+      - main
+    paths-ignore: # This expression needs to match the paths ignored on test.yml.
+      - '**'
+      - '!**/*.md'
+  pull_request:
+    paths-ignore: # This expression needs to match the paths ignored on test.yml.
+      - '**'
+      - '!**/*.md'
```
amannocci, line 16:
```suggestion
    paths: # This expression needs to match the paths ignored on test.yml.
      - '**.md'
```
We can use the `paths` expression instead of the `paths-ignore` expression.

.github/workflows/test-docs.yml
```diff
@@ -0,0 +1,25 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  test:
+    runs-on: macos-latest
```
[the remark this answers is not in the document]
v1v: Good point, I took the current implementation in https://github.com/elastic/synthetics-recorder/blob/a6f4b6cb915d94a74a06247e379c8a141f32af34/.github/workflows/test.yml#L14 but it seems that's irrelevant for running this. Thanks

.github/workflows/test-docs.yml
v1v: I don't recall what's the reason we used `paths-ignore` but that's how we have been doing in other places:
- https://github.com/elastic/apm-agent-java/blob/5ff1430a37021be595aca1777914cfb2c3f15a04/.github/workflows/main-docs.yml#L7-L10

I'd say to keep the same functionality for now

amannocci (approved):

PR #497, opened 2024-01-25
[Spike] [DO NOT MERGE] ON Week 2024

justinkambic:
## Summary

As part of my work for ON Week 2024, I added a Project Management feature to the Recorder. This interacts with the Synthetics CLI init feature to generate a project. The user can then access the normal Recorder functionality to record a script and save it to their project.

From the Management UI, the user can push a project to Kibana as they would do via CLI. They can also delete projects, and see the status of any monitors that are running from their projects in Kibana.

## Implementation details

Adds a significant amount of backend functionality to the electron server to persist the necessary information to push to the Kibana server, interact with Synthetcs, and manage files.


tmcentee:
Not how I would do it but good job so far

justinkambic:
> Not how I would do it

Feel free to contribute to the Recorder any time! 😃

PR #534, opened 2024-10-22
fix: update playwright and synthetics to 1.47.0

vigneshshanmugam:
+ To get the necessary deps to make the next release ready.

+ Requires https://github.com/elastic/synthetics/pull/971 to generate the right code for exporting and running tests locally.


NOTICE.txt
```diff
@@ -311,219 +311,6 @@ OTHER DEALINGS IN THE SOFTWARE.
 
 
 
----
```
vigneshshanmugam, line 314: this got changed to PW, maybe we need to regenerate the notice again.

justinkambic: ++ sorry if I'm breaking stuff, trying to figure out why the electron process won't launch for e2e tests.

package.json
```diff
@@ -15,7 +15,6 @@
     "pack": "npm run build && electron-builder",
     "release": "npm run pack -- -mwl --publish=always",
     "release-ci": "npm run pack -- -mwl --publish=never",
-    "fix-sharp": "node node_modules/sharp/install/libvips && node node_modules/sharp/install/dll-copy",
```
vigneshshanmugam, line 18: We havent been using this? Usually if we dont do this, then the compiled version would be done for different envs?

vigneshshanmugam (commented):
LGTM

jasonrhodes (approved):
Approving on behalf of Vignesh and Justin

justinkambic (approved):

package.json
justinkambic: See https://github.com/lovell/sharp/issues/3750

They have overhauled their install, the script we used for this step is no longer there. Seems like it's more similar to prebuildify to add `libvips` now, but if this doesn't work we will have to revisit and figure out how to get this working again.

justinkambic (approved):

package.json
vigneshshanmugam: Nice, yeah thanks for the link. I remember doing something similar for synthetics

jasonrhodes (approved):
Approving again

andrewvc (approved):

PR #538, opened 2024-12-16
ci: fix tags regex

v1v:


justinkambic (approved):
LGTM - tested against a typical tag string using a regex test tool to verify.

PR #539, opened 2024-12-18
fix: release artifacts creation

v1v:
## Summary

Trying to fix 

image

### DIFF
--- a/.buildkite/scripts/create-github-release.sh
+++ b/.buildkite/scripts/create-github-release.sh
@@ -26,6 +26,12 @@ fi
 
 echo "--- Run GitHub release"
 if [ -n "${BUILDKITE_TAG}" ] ; then
+
+  if [ ! -d "$DIST_LOCATION" ] ; then
+    echo "No signed artifacts found in ${DIST_LOCATION}"
+    exit 1
+  fi
+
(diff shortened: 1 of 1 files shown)


.buildkite/scripts/create-github-release.sh
```diff
@@ -34,7 +40,7 @@ if [ -n "${BUILDKITE_TAG}" ] ; then
     --draft \
     --title "${BUILDKITE_TAG}" \
     --repo elastic/synthetics-recorder \
-    "${DIST_LOCATION}/*.*"
+    ./$DIST_LOCATION/*.*
```
reakaleek, line 43: NIT: without looking at the details.

`DIST_LOCATION` could be an absolute path. which would make this path wrong.

reakaleek (approved):

.buildkite/scripts/create-github-release.sh
v1v: it's not, see [L9](https://github.com/elastic/synthetics-recorder/pull/539/files#diff-275b8b0b9f438320f9038499a9ed4a17b7c185a9145ccbebcdf4cde0e0d127ffR9)

justinkambic (approved):
LGTM

PR #543, opened 2025-03-26
Upgrade EUI to v101

justinkambic:
## Summary

Adds an upgrade to the version of EUI.

Also upgrades TypeScript. **NOTE** this TS upgrade will break peer dependencies for `react-scripts`. _So far_ it does not seem like there's any truly negative impact, but we should migrate off of `react-scripts` at some point if possible and opt for a lower-footprint custom server launch experience. This work has never been explored and is likely not as daunting as it initially sounds.

## Implementation details

Simply installs some upgrades.

This does now require us to use the `--legacy-peer-deps` escape hatch for calls to `npm install` or `npm ci` because this patch will break peer deps for some older packages that refuse to support TypeScript v5+.

1. Build and run the Recorder
2. Record a journey
3. Replay the journey
4. Export the journey
5. Add some custom steps and rearrange the step dividers

### DIFF
--- a/.buildkite/scripts/package.sh
+++ b/.buildkite/scripts/package.sh
@@ -10,7 +10,7 @@ fi
 
 echo "--- Install node and gather dependencies"
 nvm install "$(cat .nvmrc)"
-npm ci
+npm ci --legacy-peer-deps
 
 echo "--- run release-ci"
 # Disable signing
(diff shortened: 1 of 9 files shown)


justinkambic:
Closing in favor of https://github.com/elastic/synthetics-recorder/pull/548.

PR #548, opened 2025-05-21
Migrate Build Process and Upgrade deps

justinkambic:
## Summary

Lots of goodness here.

1. Migrate to a custom webpack config. This enables us to delete `create-react-app` from the repo, which also means we can upgrade peer deps like TypeScript, which we were previously locked into.
2. This bumps a bunch of dependencies, and lets us have future flexibility to more easily address module-level security issues when they arise.
3. Modern EUI with Borealis theme, which will give our users access to a look and style that is consistent with the Kibana UIs they are likely used to.
4. Migrates component-level style reqs from `styled-components` to `@emotion/react`; we are able to drop `styled-components` as a dep altogether and further unify the recorder's codebase with the conventions in Kibana.
5. Lazy loading React components in the top-level app file. The benefit is likely marginal, but this is a best practice we should follow and a precedent we should set for future contributions.
6. Improved DX with better source maps; devs can now walk through the actual TypeScript code in the Recorder's DevTools.
7. More build flexibility. We have more top-level deps than before, but it's because we removed `create-react-app` and kept all the features we need for our specific build.

## Implementation details

There shouldn't be any user-facing impact, aside from the updated UI look and feel, and perhaps some minor layout changes.

Ensure all core features continue to work. This means you can record journeys, overwrite them, export them as code, and test them against Playwright's Chromium.


justinkambic:
This would be good to merge in the future. If someone wants, re-open it and review it.

PR #578, opened 2025-12-11
bk: notify to a different slack channel

v1v:
## Summary
Change the slack channel to one that it's accessible by the synthetics team


fr4nc1sc0-r4m0n (approved):
LGTM

PR #587, opened 2025-12-22
docs(download): update download docs

cesco-f:
Update Download docs


justinkambic (approved):
LGTM, download links all work, date/version is correct. 🚢

PR #588, opened 2026-04-13
Bump @elastic/synthetics to 1.23.2

shahzad31:
## Summary

This change upgrades `@elastic/synthetics` from **1.22.0** to **1.23.2** and refreshes `package-lock.json` so transitive dependencies resolve to current releases.

## Motivation

Newer `@elastic/synthetics` pulls updated dependency ranges (notably `undici` `^7.18.2`), which resolves to a current `undici` 7.x line in the lockfile instead of older pinned minors. That addresses several high-severity findings tied to the previous `undici` resolution without using npm `overrides`.

## Changes

- `package.json`: bump `@elastic/synthetics` to `1.23.2`.
- `package-lock.json`: regenerated for the new version and its dependency tree.

## Notes

- Playwright packages in this repo remain the custom **1.56.1-recorder** tarballs; `@elastic/synthetics` 1.23.2 still targets Playwright **1.56.1**, which aligns with that pin.
- Other vulnerability findings that come from unrelated dev or build tooling (for example `react-scripts`, Electron tooling, or nested cache/glob stacks) are outside this dependency bump and may still need separate follow-up.

## Testing

- [ ] `npm install` (already run locally when updating the lockfile)
- [ ] `npm run test:unit` / CI green

Made with [Cursor](https://cursor.com)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
       "dependencies": {
         "@elastic/datemath": "^5.0.3",
         "@elastic/eui": "^89.1.0",
-        "@elastic/synthetics": "1.23.2",
+        "@elastic/synthetics": "=1.23.2",
         "@emotion/cache": "^11.9.3",
         "@emotion/react": "^11.9.3",
         "archiver-utils": "^5.0.2",
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
       "dependencies": {
         "@elastic/datemath": "^5.0.3",
         "@elastic/eui": "^89.1.0",
-        "@elastic/synthetics": "=1.22.0",
+        "@elastic/synthetics": "1.23.2",
         "@emotion/cache": "^11.9.3",
         "@emotion/react": "^11.9.3",
         "archiver-utils": "^5.0.2",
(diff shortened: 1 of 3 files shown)


miguelmartin-elastic (dismissed):

package.json
```diff
@@ -87,7 +87,7 @@
   "dependencies": {
     "@elastic/datemath": "^5.0.3",
     "@elastic/eui": "^89.1.0",
-    "@elastic/synthetics": "=1.22.0",
+    "@elastic/synthetics": "1.23.2",
```
shahzad31:
```suggestion
    "@elastic/synthetics": "=1.23.2",
```

package-lock.json
```diff
@@ -12,7 +12,7 @@
       "dependencies": {
         "@elastic/datemath": "^5.0.3",
         "@elastic/eui": "^89.1.0",
-        "@elastic/synthetics": "=1.22.0",
+        "@elastic/synthetics": "1.23.2",
```
shahzad31:
```suggestion
        "@elastic/synthetics": "=1.23.2",
```

miguelmartin-elastic (approved):

